PR #7, opened 2025-10-16
Fix all compiler warnings

quad:
The objc bindings deprecated official symbol names for Rust-ified equivalents. In the process, many formerly unsafe functions are now safe.


karinushka:
Thanks a lot!
I realized that my local cargo repo wasn't synced for quite some time, so I did not have all these warnings because I was building against the older objc bindings.

PR #8, opened 2025-10-16
Set MSRV to 1.88.0

quad:
if-let-chain as seen in src/platform.rs:684:16


quad:
IMHO it's not an issue; stick with the Rust _you_ use and love! In my anecdotal experience everyone runs stable or nightly via `rustup` or a toolchain manager like `mise`. It's not a significant bar to contribution to require stable.

PR #15, opened 2025-10-17
Set the Rust toolchain version to stable

quad:
Related to #8. This causes `rustup` managed binaries (`cargo`, …) to automatically use the latest stable version of Rust when ran in this repository.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,3 @@
+[toolchain]
+channel = "stable"
+profile = "default"


alperchoco:
I got a weird error and then had to figure out that this compiles on 1.88. Also couldn't find a release target?

quad:
@rami3l wow I have to take this opportunity to ask the authority: what is the best way to do this sort of thing?

rami3l:
@quad I think it depends largely on your specific configuration? Without knowing more about your situation I'd say if this were my project I might recommend using `rustup override set` to configure a local preference on your side rather than enforcing that to everyone :)

PR #20, opened 2025-10-18
feat: implement daemon lifecycle management

### LINKED ISSUE #5: Document and/or enhance lifecycle management of the daemon
Many thanks for making this project, and it looks quite promising!

There is one question regarding the installation instructions though: assuming that you are daily-driving this project, how are you managing this window manager once compiled? Do you put it directly into "Login Items" in system settings to start the daemon, and use the "quit" hotkey to stop it? What would be the recommended practice?

Many thanks in advance!

rami3l:
Closes #5.

As previously discussed, this PR uses copied and slightly modified code from https://github.com/rami3l/clavy, removing some `clavy`-specific logic.

Tested locally and seems to work OOTB with the following commands (might need to grant accessibility permissions): 

```console
> cargo install --path .
> paneru install
> paneru start
```

This comment is written on my own Mac in that very state.

## Key changes

* The log output target has been switch back to `stdio`, since `launchctl` will handle that for us from now on.

* New subcommands such as `paneru install` and `paneru start` have been introduced for basic daemon lifecycle control. The original behavior is covered by `paneru launch` which `paneru` with no additional args falls back to.
  ```console
  > paneru --help

  Usage: paneru [COMMAND]

  Commands:
    launch     Launch the daemon directly in the console
    install    Install the service
    uninstall  Uninstall the service
    reinstall  Reinstall the service
    start      Start the service
    stop       Stop the service
    restart    Restart the service
    help       Print this message or the help of the given subcommand(s)

  Options:
    -h, --help     Print help
    -V, --version  Print version
  ```

* The argument parsing logic has been slightly changed: the `paneru send-cmd` subcommand (hidden from the help string) has been introduced for the old behavior where arg count is larger than 1.

## Concerns

It is out of the scope of this PR, but I suggest adding an additional mechanism for checking AX privileges [like so](https://github.com/koekeishiya/yabai/blob/a8eb6b1a7da4e33954b716b424eb51ce47317865/src/misc/helpers.h#L328)...

Otherwise it might trigger a macOS bug when the binary is updated which will cause the permission to be invalid, and you will need to go to system settings to remove and re-grant permissions to `paneru` to make things work again.

Ideally checking for once on startup would be enough.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -231,6 +231,6 @@ fn launch() -> Result<()> {
 
     platform_callbacks.run(quit);
 
-    handle.join().expect("Can not joing threads at the end.");
+    handle.join().expect("Cannot join threads at the end.");
     Ok(())
 }


assets/launchd.plist
```diff
@@ -0,0 +1,32 @@
...
+    <dict>
+      <key>NO_COLOR</key>
+      <string>1</string>
+    </dict>
+    <key>RunAtLoad</key>
+    <true />
+    <key>StandardErrorPath</key>
+    <string>{out_log_path}</string>
+    <key>StandardOutPath</key>
+    <string>{error_log_path}</string>
```
rami3l: The convention used in `clavy` is that only severe error messages and friends will be printed to stdout, thus the stdout goes to `.err` and the stderr goes to `.out`. However you are free to adjust this (e.g. revert this) and even disable one of the two if you consider it unnecessary.

assets/Info.plist
```diff
@@ -0,0 +1,14 @@
...
+<plist version="1.0">
+  <dict>
+    <key>CFBundleIdentifier</key>
+    <string>com.github.karinushka.paneru</string>
+    <key>CFBundleShortVersionString</key>
+    <string>0.1.0</string>
+    <key>CFBundlePackageType</key>
+    <string>BNDL</string>
+    <key>NSHumanReadableCopyright</key>
+    <string>Copyright (c) 2025 Karinushka@github. All rights reserved.</string>
```
rami3l: I have copied what is in the LICENSE over here. Hope that's okay for you.

src/main.rs
```diff
@@ -13,10 +13,13 @@ mod events;
 mod manager;
 mod platform;
 mod process;
+mod service;
 mod skylight;
 mod util;
 mod windows;
 
+embed_plist::embed_info_plist!("../assets/Info.plist");
```
rami3l: This is necessary in the (upcoming) distribution phase (via e.g. Homebrew) since a distributed binary has to built in a specific way in order to use AX privileges. Embedding `Info.plist` should be enough to make it so.

See https://crates.io/crates/embed_plist#motivation for what this is doing.

Cargo.toml
```diff
@@ -7,7 +7,10 @@ rust-version = "1.88.0"
 [dependencies]
 accessibility-sys = "0.1.3"
 chrono = "0.4.42"
+embed_plist = "1.2.2"
```
rami3l: Question: Why was `Cargo.lock` excluded?

PR #23, opened 2025-10-20
chore: Allow config to disable swipe

seivan:
Since the configuration will automatically fail if a key is missing, I'm forced to add the key as optional, otherwise this would break it for a lot of users. 

My suggestion is that the configuration should load defaults if keys are missing, hell the entire configuration should load a default if missing, but that should be a different PR. 

I use Swipes for historical navigation, like in Safari or Finder, so I need it disabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ starting point:
 [options]
 # Enables focus follows mouse
 focus_follows_mouse = true
+swipe_enabled = true
 
 [bindings]
 # Moves the focus between windows.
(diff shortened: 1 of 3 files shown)


karinushka:
> And yeah `swipe_gesture_fingers = 4` does not disable it for `2` or `3` fingers.

Super weird! What touchpad are you using? The fact that it triggers the gesture with just 2 points is strange - the code will ignore anything lower than 3. So it must be the device or the OS which somehow artificially reports a swipe. What MacOS version do you have?

I am using the built-in macbook touchpad and an external magic trackpad when docked. And it works correctly for both cases - independent tracking of 3, 4 and 5 (even 6) fingers.

PR #46, opened 2025-12-05
Fix focus issues

kremovtort:
While I was using **paneru**, I encountered a serious problem – if I close or hide the window, then there is an empty space in place of the window, through which `window_focus_(west|east)` keybindings cannot be moved, in this PR I make several changes to solve this problem (unfortunately not completely):
1. Add to event handling `WindowDestroyed` and `WindowMinimized` cases, when the destroyed/minimized window was focused and I move the focus to the neighbor window.
2. For `WindowMinimized`, I also add `ReshuffleAroundTrigger`.
3. If the window on the left cannot be found with `WindowDestroyed` and `WindowMinimized`, then we look for the window on the right for `ReshuffleAroundTrigger`.
4. For commands to change focus, I add the handling of the case when the window to focus is already hidden.

There is still a problem with the fact that many applications, when closing a window, do not actually call `WindowDestroyed`, but only move it outside the display. Attempts to fix this caused me a problem with the fact that the windows that **paneru** controls and were moved by it outside the display were also removed from **paneru**. I tried several ways to fix this, but it didn't work out, so I decided to send this PR, because it already makes the user experience much better for many cases.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -126,18 +126,9 @@ impl WindowManager {
 
                 // Handle focus transfer and reshuffle similar to window destruction
                 if let Ok(panel) = active_display.active_panel(main_cid) {
-                    // Try to find a window to the left first
-                    _ = panel.access_left_of(entity, |p| {
+                    _ = panel.access_neighbor_of(entity, |p| {
                         neighbor_entity = p.top();
-                        false
                     });
-                    // If no left neighbor, try to find a window to the right
-                    if neighbor_entity.is_none() {
-                        _ = panel.access_right_of(entity, |p| {
-                            neighbor_entity = p.top();
-                            false
-                        });
-                    }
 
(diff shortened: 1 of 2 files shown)


karinushka:
Check out this fix on the `testing` branch: https://github.com/karinushka/paneru/commit/6a753a8c8de2d7f939822d655627901b2f90021b
Does it work for you? It is simpler to just issue a `WindowFocused` trigger and let the existing handler move the focus marker.

kremovtort:
> Check out this fix on the `testing` branch: [6a753a8](https://github.com/karinushka/paneru/commit/6a753a8c8de2d7f939822d655627901b2f90021b) Does it work for you?

Well it doesn't works perfectly. I still can focus hidden application and then I can't focus another window with paneru command.

Also closing applications like telegram and paneru broke pane so there is empty space that I can't move through with paneru command.

PR #67, opened 2026-01-17
chore(nix): make HM settings optional

postsolar:
Default settings to null and only write the config file when provided.

Why: currently the default for settings is `{}` and the only check on whether to write the config is whether the service is enabled. This doesn't let a user enable the service and *not* create a config via the module.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -54,8 +54,8 @@
             };
 
             settings = lib.mkOption {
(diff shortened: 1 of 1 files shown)


postsolar:
I had the same PR here https://github.com/JakeStanger/ironbar/pull/1046

PR #81, opened 2026-02-08
feat: Implement Niri-style viewport panning for trackpad swipe gestures

zzzhizhia:
## Summary

- Replace the old swipe behavior (which moved the focused window within the strip and switched focus at edges) with Niri-style whole-strip viewport panning: all windows move together like a camera panning across a long desk
- Directly reposition windows via AX API during swipe for immediate response, skipping the animation system
- Skip AX API calls for off-screen windows during panning, only updating internal coordinates for performance
- Separate swipe and non-swipe layout paths: swipe uses unclamped offsets for free panning, while window close/focus change retains per-window clamping to keep windows tightly packed

## Test plan

- [x] `cargo test` — 8/8 tests pass
- [ ] Open 3+ windows, swipe trackpad to verify all windows pan together smoothly
- [ ] Close a window, verify remaining windows collapse with no gaps
- [ ] Verify keyboard shortcuts (focus/swap west/east) still work correctly
- [ ] Verify clicking a window at screen edge still scrolls viewport to it

### DIFF
--- a/src/ecs/systems.rs
+++ b/src/ecs/systems.rs
@@ -793,19 +793,22 @@ pub(super) fn window_swiper(
             })
         };
 
-        let Some(positions) = calculate_positions(
-            entity,
-            new_x,
-            active_strip,
-            &window_width,
-        ) else {
+        // 计算纯偏移位置（无钳制），用于滑动平移
+        let abs_positions: Vec<f64> = absolute_positions(active_strip, &window_width).collect();
+        let Some(&anchor_abs) = active_strip
+            .index_of(entity)
+            .ok()
+            .and_then(|index| abs_positions.get(index))
+        else {
             continue;
         };
+        let offset = new_x - anchor_abs;
 
-        let targets: Vec<_> = positions
+        let targets: Vec<_> = abs_positions
+            .into_iter()
             .zip(active_strip.all_columns())
-            .filter_map(|(position, col_entity)| {
-                window_width(col_entity).map(|_| (position, col_entity))
+            .filter_map(|(abs_pos, col_entity)| {
+                window_width(col_entity).map(|_| (abs_pos + offset, col_entity))
             })
             .collect();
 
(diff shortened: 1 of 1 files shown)


karinushka:
Thanks for checking this out! The swiping functionality was somewhat crufty and needed a cleanup.
There is no need cull off-screen movements or have `set_origin()` method - the check already happens in the `reposition()` function: https://github.com/karinushka/paneru/blob/66c94493cc99ee7715b4beb1e5cf455ada60213a/src/manager/windows.rs#L368-L371

I simplified the PR slightly and also removed the old `continuous_swipe` option, because it has been on by default and there are no obvious benefits of having it turned off - https://github.com/karinushka/paneru/commit/0e35ced1b8ecbf08768d982fc811206ece2adbe1 on the `testing` branch.
Please check it out and see whether everything works correctly for you.

src/ecs/systems.rs
```diff
@@ -765,75 +765,68 @@ pub(super) fn animate_resize_windows(
...
-    for (window, entity, WindowSwipeMarker(delta)) in sliding {
+    let active_strip = active_display.active_strip();
+    let display_bounds = active_display.bounds();
+
+    for (entity, WindowSwipeMarker(delta)) in sliding {
         commands.entity(entity).try_remove::<WindowSwipeMarker>();
 
-        let pos_x = window.frame().origin.x - (active_display.bounds().size.width * delta);
-        let frame = window.frame();
+        let display_width = display_bounds.size.width;
```
karinushka: Move this assignment outside of the loop - maybe to line 775.

karinushka (changes requested):
Added a couple of small suggestions - please take a look.

zzzhizhia:
Thanks for your review! I’ll test it out as soon as I’m free.

karinushka:
Check out the latest `testing` branch commit https://github.com/karinushka/paneru/commit/b413014b7ff300a7809d9bc508e0c16757e8ebcf
Yet again the scrolling was reworked and is now extra smoooooth. Give it a try.

PR #99, opened 2026-02-24
feat: add pop-out transition when toggling window unmanaged

skokenes:
## Summary
- Add visible pop-out behavior when a window becomes `Unmanaged::Floating` in `window_unmanaged_trigger`.
- Cap the transitioning window size to at most 80% of display width and height.
- Apply a small on-screen offset so the window clearly appears outside the managed strip without leaving screen bounds.

## Why
- Toggling unmanaged previously gave little visual feedback if geometry stayed the same.
- This makes the state change obvious while keeping the window usable and visible.

## What Changed
- `src/ecs/triggers.rs`
  - Added transition constants for max-size ratio (`4/5`) and pop offset (`32`).
  - Added clamping helper to keep the target frame within display bounds.
  - Added offset candidate logic that only applies in-bounds moves.
  - On unmanaged transition, shrink width/height only when they exceed 80% of display size, then reposition with safe offset.

## Verification
- `cargo build`
- `cargo test` (11 passed, 0 failed)

## Notes
- This is focused on the unmanaged transition behavior in triggers and keeps the existing strip reshuffle behavior intact.

### DIFF
--- a/src/ecs/triggers.rs
+++ b/src/ecs/triggers.rs
@@ -22,12 +22,13 @@ use crate::config::{Config, WindowParams};
 use crate::ecs::params::{ActiveDisplay, ActiveDisplayMut, Configuration, Windows};
 use crate::ecs::{
     ActiveWorkspaceMarker, LocateDockTrigger, SendMessageTrigger, WindowSwipeMarker,
-    reposition_entity, reshuffle_around,
+    reposition_entity, reshuffle_around, resize_entity,
 };
 use crate::errors::Result;
 use crate::events::Event;
 use crate::manager::{
-    Application, Display, LayoutStrip, Process, Window, WindowManager, WindowPadding, irect_from,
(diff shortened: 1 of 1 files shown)


src/ecs/systems.rs
```diff
@@ -724,6 +729,168 @@ pub(super) fn animate_resize_windows(
...
+        let Some((display, dock)) = displays
+            .iter()
+            .find(|(display, _)| display.id() == marker.display_id)
+        else {
+            warn!(
+                "Display {} not found for deferred reposition of window {}",
+                marker.display_id,
+                window.id()
+            );
+            commands.entity(entity).try_remove::<DeferredRepositionMarker>();
```
karinushka: This marker is removed three times - once for each outcome. It is simpler to just remove it once at the start of the for loop, that way it will not be forgotten later on.

PR #147, opened 2026-03-25
fix: prevent focus lock during rapid keyboard shortcut switching

Fjx-dylanZ:
## Summary

  Fixes a race condition where rapidly pressing focus keyboard shortcuts could lock
  focus onto one window, requiring a mouse click to unstick it. Three related issues:

  - **Stale focus events corrupt FocusedMarker**: `window_focused_trigger` transferred
    `FocusedMarker` unconditionally, then checked `is_frontmost()` — too late. Delayed
    events (from `RetryFrontSwitch`, `dont_focus` re-assertions, etc.) could pull focus
    back to an old window, desynchronizing Paneru from macOS. Moved the `is_frontmost()`
    guard before the transfer and added a `focused_window_id()` check to also catch
    same-app stale events.

  - **Rapid keypresses swallowed**: `command_move_focus` relied on the OS event
    round-trip to update `FocusedMarker`. Multiple presses before the round-trip
    completed would all target the same window. Now updates `FocusedMarker` immediately.

  - **Stale `RetryFrontSwitch`**: A transient AX error during a fast cross-app switch
    spawned a retry that could fire seconds later for a no-longer-frontmost app. Added
    an `is_frontmost()` guard in `retry_front_switch`.

  ## Tests (created via Claude Code)

  - [x] Two new regression tests that reproduce both race conditions
    - `test_rapid_focus_not_swallowed` — three focus commands across consecutive frames
      without event flushing; verifies each press advances one step
    - `test_stale_focus_event_ignored` — injects a stale `WindowFocused` event after
      focus has moved; verifies `FocusedMarker` is not pulled back
  - [x] Both tests fail on `origin/main`, pass with these changes
  - [x] All 42 tests pass
  - [x] Manual testing: rapid focus switching no longer locks

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -1526,3 +1526,121 @@ fn test_window_hidden_ratio_swap() {
 
     run_main_loop(&mut bevy, &internal_queue, &commands, check);
 }
+
+/// Verify that focus state is on the expected window.
+fn verify_focused_window(expected_id: WinID, world: &mut World) {
+    let mut query = world.query::<(&Window, Has<FocusedMarker>)>();
+    let focused: Vec<_> = query
+        .iter(world)
+        .filter(|(_, focused)| *focused)
+        .collect();
+    assert_eq!(focused.len(), 1, "expected exactly one focused window");
+    assert_eq!(
+        focused[0].0.id(),
+        expected_id,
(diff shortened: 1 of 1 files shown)


karinushka:
Thanks for tracking this down!

Issue #119, opened 2026-03-02
(Feature Request): Workspaces


peteqian:
I really enjoyed the window scrolling configuration but I have an issue with my workflow, it's that I use workspaces a lot, especially with Niri. 

Is there any chance that this is currently being worked on or planned in the future?

karinushka:
Hi folks
Check out this virtual workspaces commit in the latest `testing` branch: https://github.com/karinushka/paneru/commit/56117be122a5eecfd3f9213eb2bac28f96951b2c

Documentation for key bindings: https://github.com/karinushka/paneru/blob/testing/CONFIGURATION.md#virtual-workspaces-experimental

**TLDR;**
```toml
[bindings]
window_virtual_north = "cmd + shift - k"
window_virtual_south = "cmd + shift - j"
window_virtualmove_north = "cmd + alt - k"
window_virtualmove_south = "cmd + alt - j"
```

*Note:* for now it hides the strips at the bottom edge of the screen - this may result in weird stuff if you have a secondary monitor there.

nikosdouvlis:
Found another edge case while testing virtual workspaces. If an app has windows on multiple virtual workspaces and you close one, macOS auto-focuses another window of the same app on a different workspace. Paneru's `activate_virtual_row_trigger` follows that focus change and teleports you to the wrong workspace.

Repro:
- ws1: Chrome, Wezterm, Slack
- ws2: Spotify, Wezterm
- Focus Slack on ws1, switch to ws2, focus Wezterm, close Wezterm
- Expected: stay on ws2 with Spotify focused
- Actual: teleported to ws1

I think the issue is that `give_away_focus` only fires an internal ECS event but doesnt actually call `focus_with_raise` on the macOS side, so the OS-level focus change wins the race.

agmitron:
Looks great! I think I don't even need native "Spaces" anymore, at least for a single display :) 

> there is also a small tweak which flashes the current workspace when switching them. This should improve navigation

Is it removable? I'd prefer indicating it in the menu bar or in Sketchybar.

MaKhandare:
Love the virtual workspaces, thank you!

Is it planned to add bindings to instantly switch to workspace x, as mentioned by @imrellx, instead of cycling through them via `window_virtual_north / _south`?

karinushka:
> Is it planned to add bindings to instantly switch to workspace x

There is a PR in flight (https://github.com/karinushka/paneru/pull/204) which adds this feature - switching directly to a numeric virtual workspace.

PR #162, opened 2026-03-31
feat: persist and restore focused window per virtual workspace

nikosdouvlis:
Addresses feedback from #119 (virtual workspaces) - specifically the non-deterministic focus when switching between virtual workspaces and back.

When switching virtual workspaces and coming back (e.g. ws1 -> ws2 -> ws1), the window that gets focused is non-deterministic. In my testing, its either the first or the last window of the strip but I couldnt find a consistent pattern.

## How it worked before

If I understand correctly, `show_active_workspace_trigger` picks the first window in column order whose animated x position is past the display's left edge:

```rust
strip.all_windows().into_iter().find(|entity| {
    windows.moving_frame(*entity)
        .is_some_and(|frame| frame.min.x >= bounds.min.x)
})
```

The problem is that both `show_active_workspace_trigger` and `workspace_activated_trigger` fire on `On<Add, ActiveWorkspaceMarker>`, and `workspace_activated_trigger` does a `reshuffle_around` that can change the scroll position. The layout systems (`position_layout_strips`, `position_layout_windows`) run later in Update, not in the triggers. So `moving_frame` returns whatever the animation state happens to be at that moment, and the result depends on Bevy observer ordering and whether layout has caught up. On top of that, `workspace_activated_trigger` has its own fallback to `strip.get(0)` (column 0) which can fight the other trigger's selection.

## What this changes

This adds a `last_focused_window` field to `LayoutStrip` that remembers which window was focused on each virtual workspace. When switching back, that window is restored instead of falling back to geometry.

- `activate_virtual_row_trigger` records the focused entity on its strip whenever `FocusedMarker` is added, so it catches keyboard commands, focus-follows-mouse, cmd+tab, everything
- `show_active_workspace_trigger` prefers the stored window when activating a workspace, falling back to the first column if the entity was removed
- `workspace_activated_trigger` uses the same memory for viewport centering
- `handle_virtual_window_moves` marks the moved window as last-focused on the destination strip
- Stale references are cleared in `LayoutStrip::remove()` and validated with `index_of` at every consumption point

One bonus: since focus now lands on the right window immediately, theres no scroll animation when switching back to a workspace. Before, it would often focus the wrong window first and then scroll to re-center, which was disorienting on quick switches.

Also added a test that moves two windows to a virtual workspace, focuses one specifically, switches away and back, and verifies the right window is restored (not the first-column fallback).

Im very open to redoing this if you think a different direction would be better, just let me know :)

### DIFF
--- a/src/ecs/layout.rs
+++ b/src/ecs/layout.rs
@@ -127,6 +127,7 @@ pub struct LayoutStrip {
     id: WorkspaceId,
     pub virtual_index: u32,
     columns: VecDeque<Column>,
+    last_focused_window: Option<Entity>,
 }
 
 impl LayoutStrip {
@@ -135,6 +136,7 @@ impl LayoutStrip {
             id,
             virtual_index,
             columns: VecDeque::new(),
+            last_focused_window: None,
         }
     }
 
@@ -278,6 +280,10 @@ impl LayoutStrip {
                     }
                 }
             }
+        }
+
+        if self.last_focused_window == Some(entity) {
+            self.last_focused_window = None;
         }
     }
 
@@ -476,6 +482,19 @@ impl LayoutStrip {
 
     pub fn id(&self) -> WorkspaceId {
         self.id
+    }
+
+    pub fn validated_last_focused_window(&self) -> Option<Entity> {
+        self.last_focused_window
+            .filter(|entity| self.index_of(*entity).is_ok())
+    }
+
+    pub fn set_last_focused_window(&mut self, entity: Entity) {
+        debug_assert!(
+            self.index_of(entity).is_ok(),
+            "set_last_focused_window called with entity not in strip"
+        );
+        self.last_focused_window = Some(entity);
     }
 
     #[instrument(level = Level::TRACE, skip_all, fields(offset))]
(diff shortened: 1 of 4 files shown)


karinushka:
The way I pushed it in the current `testing` head, I save the actual strip position instead of a focused window.
https://github.com/karinushka/paneru/blob/69417f3c1f263240ed730e2fe383b50d51225f50/src/ecs/workspace.rs#L31-L32
This should be more robust - because a window can despawn or get moved, while a virtual strip position is quite stable.

The only issue in the current `test` branch is that the saved position is slowly crawling to the side - I think it's related to he race condition you mention - e.g. two triggers activating at the same time. I pushed the current state to testing to avoid further rebasings, but will investigate it in the meanwhile.
Maybe using a `RemovedComponents<>` can help:
https://github.com/karinushka/paneru/blob/69417f3c1f263240ed730e2fe383b50d51225f50/src/ecs/workspace.rs#L754-L756

nikosdouvlis:
> Check out this latest commit in the `testing` branch: [2e4742f](https://github.com/karinushka/paneru/commit/2e4742f89df8fb096c8faad3fc2ddb113656a5d7) the solution ended up a bit more complicated, but I ironed out a lot of edgecases.

My initial tests look great and I think you also eliminated all flickering I was observing while switching from one monitor to another - I can provide more details about this if you need of course but `testing` has been super stable since yesterday!

PR #167, opened 2026-04-01
feat: vertical gesture navigation for virtual workspaces

nikosdouvlis:
Paneru supports vertical stacking of virtual workspaces (multiple workspace rows per native macOS space), but navigating between them was keyboard-only. This adds 3/4-finger trackpad gesture support so vertical swipes switch between virtual rows, matching how horizontal swipes already scroll the strip.

The gesture works by firing the existing `window virtual north` / `window virtual south` commands, so theres no new workspace switching logic. The only new code is axis detection in the input layer and a small ECS system that accumulates the gesture delta and triggers the command at threshold.

How it works:

- `handle_swipe` now reads both X and Y from `normalizedPosition()` and routes by axis dominance. Horizontal goes to the existing `Swipe` event, vertical goes to a new `VerticalSwipe` event
- A new `vertical_swipe_gesture` ECS system accumulates vertical delta across frames and fires `Operation::Virtual(North/South)` when it crosses a threshold. It fires once per gesture and resets when fingers lift
- Scroll wheel also supports vertical switching through an optional `vertical_modifier` config (e.g. alt+shift+scroll = switch rows, while alt+scroll = horizontal strip scroll as before)
- Scroll events are suppressed for 500ms after a vertical gesture to prevent macOS momentum scroll from scrolling content inside the destination workspace

Config example:

```toml
[swipe.scroll]
modifier = "alt"
vertical_modifier = "shift"
```

Ref: https://github.com/karinushka/paneru/issues/119#issuecomment-4159016193

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -61,6 +61,7 @@ Configure trackpad gestures and scroll-wheel window sliding.
 | Option | Type | Default | Description |
 | :--- | :--- | :--- | :--- |
 | `modifier` | String | `"alt"` | Modifier key(s) required to slide windows with the scroll wheel: `"alt"`, `"rcmd"`, `"ralt + cmd"`, `"lctrl + lalt + cmd"`, etc. |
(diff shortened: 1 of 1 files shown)


karinushka:
This looks cool! Thanks a lot.
Unfortunately I force pushed the testing, so you may need to rebase. Just sync to top of `testing` again - I promise to chill with force-pushes in the meanwhile.

PR #177, opened 2026-04-07
feat(nix): add nix-darwin module, test & refactor using flake-parts

emonadeo:
## TLDR
```
- feat: adds a nix-darwin module
- test: nix-darwin module (home-manager module is still untested though)
- refactor: split flake.nix into submodules inside /nix folder using flake-parts
- refactor: rename launchd agent from `Paneru` to `com.github.karinushka.paneru`
- docs: rework nix docs and move them out of README
- chore: update flake inputs
```

This is generally<sup>1</sup> a non-breaking change.

<sup>1</sup>The only thing that changes for flake users is that the launchd agent is no longer called `Paneru` but `com.github.karinushka.paneru` to match the behavior of the `paneru install` cli command and naming convetions. It shouldn't break existing home-manager configurations.

### New nix-darwin module

I recently stopped using home-manager but still want to use paneru.
This PR first and foremost adds a darwin module that is used the exact same way as the home module:

```nix
# configuration.nix
{ inputs, ... }:
{
  imports = [
    inputs.paneru.darwinModules.paneru
  ];

  services.paneru = {
    enable = true;
    settings = {
      options = {
        focus_follows_mouse = true;
        preset_column_widths = [0.25 0.33 0.5 0.66 0.75];
      };
      # ...
    };
  };
}
```
The implementation is also very similar to the existing home-module with one key difference:
Instead of generating a `~/.paneru.toml` or `$XDG_CONFIG_HOME/paneru/paneru.toml` the configuration is written into the nix store and linked using `$PANERU_HOME`. Zero home directory pollution yay.

### Tests

`nix/checks.nix` contains an integration test that checks if the launchd agent and paneru configuration files are generated properly.

The test(s) can be run with `nix flake check`.

They are akin to how nix-darwin internally tests its modules:
- <https://github.com/nix-darwin/nix-darwin/blob/master/release.nix>
- <https://github.com/nix-darwin/nix-darwin/issues/1227>

### Refactor

With the darwin module and tests the flake.nix got really big, so using [flake-parts](https://github.com/hercules-ci/flake-parts) I split these into different files that should be simple enough to understand their purpose.

---

This PR got kinda big, sorry about this (zero AI-generated code I swear)


emonadeo:
@karinushka I see you just merged it. If any issues arise related to Nix where you might need help on feel free to mention me 👍

PR #180, opened 2026-04-08
fix: Update passthrough keybindings before early-return in window_foc…

CyrilPeponnet:
…used_trigger

When focus moves via keyboard (focus_entity), FocusedMarker is set before the OS WindowFocused event arrives. The early-return guard in window_focused_trigger then skips update_passthrough, leaving stale passthrough keys from a previously focused window.

This causes keybindings to be silently swallowed: e.g. after focusing away from a window with bindings_passthrough (like neovide with ctrl-h/l), the passthrough remains active on the newly focused window, preventing paneru from intercepting those keys.

Move update_passthrough before the early-return so it always runs when the OS confirms focus, regardless of whether FocusedMarker is already set.

### DIFF
--- a/src/ecs/triggers.rs
+++ b/src/ecs/triggers.rs
@@ -231,12 +231,6 @@ pub(super) fn window_focused_trigger(
         return;
     };
 
-    if let Some((window, _)) = windows.focused()
-        && window.id() == window_id
-    {
-        return;
-    }
-
     let Some((window, entity, parent)) = windows.find_parent(window_id) else {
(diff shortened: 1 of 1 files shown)


src/ecs/triggers.rs
```diff
@@ -248,6 +242,18 @@ pub(super) fn window_focused_trigger(
         return;
     };
 
+    // Always keep passthrough in sync, even when the ECS already considers
```
karinushka: I think this block should be above the StrayFocusEvent check above (L234-243).
That way the logic is the same - e.g. if window is not focused, it will launch the stray focus marker.

karinushka (changes requested):

PR #185, opened 2026-04-14
fix: disable AXEnhancedUserInterface during window move/resize

### LINKED ISSUE #175: Firefox / zen etc animation drag behind
When focusing firefox the animation/movement of it doesnt start until all other windows has repositioned. Both for `window focus` and `window virtual...`

It gets very annoying because it shows over/under the other windows when they move (and then when firefox moves as well) and when going to a virtual space it comes in/hides from the corner in a visible animation.

I recorded it with very low `animation_speed` so its easier to see:

https://github.com/user-attachments/assets/4b4ec69c-fbea-41a8-8576-98e114f3847a

I had clanker try to fix it, but it failed and said it might be because of how firefox handles the accessibility apis? :/

Fjx-dylanZ:
Temporarily disable the attribute before each reposition/resize call and restore it afterward, using a per-PID ref-count for par_iter_mut safety. This is the same workaround used by yabai and Hammerspoon.

https://github.com/asmvik/yabai/blob/5e213518d413d2bc41159a759d0e7de3adf9d799/src/misc/helpers.h#L524-L531

https://github.com/Hammerspoon/hammerspoon/blob/08e93f679bb5d9b88d2e8bd493d964a133c89960/Hammerspoon/HSuicore.m#L748-L782

this should resolve #175

### DIFF
--- a/src/manager/windows.rs
+++ b/src/manager/windows.rs
@@ -1,17 +1,20 @@
 use accessibility_sys::{
-    AXUIElementRef, AXValueCreate, AXValueGetValue, kAXFloatingWindowSubrole, kAXPositionAttribute,
-    kAXRaiseAction, kAXSizeAttribute, kAXStandardWindowSubrole, kAXUnknownSubrole,
-    kAXValueTypeCGPoint, kAXValueTypeCGSize, kAXWindowRole,
+    AXUIElementCreateApplication, AXUIElementRef, AXValueCreate, AXValueGetValue,
+    kAXFloatingWindowSubrole, kAXPositionAttribute, kAXRaiseAction, kAXSizeAttribute,
+    kAXStandardWindowSubrole, kAXUnknownSubrole, kAXValueTypeCGPoint, kAXValueTypeCGSize,
+    kAXWindowRole,
 };
 use bevy::ecs::component::Component;
 use bevy::math::IRect;
 use core::ptr::NonNull;
 use derive_more::{DerefMut, with_trait::Deref};
 use objc2_core_foundation::{
-    CFArray, CFNumber, CFRetained, CFString, CFType, CGPoint, CGRect, CGSize,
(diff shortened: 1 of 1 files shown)


karinushka:
Thanks for the fix! 😺

I adjusted it slightly: https://github.com/karinushka/paneru/commit/7598a0ecc34d9ed3fb31cbb30222fc2bc1d57e3e
Basically caching the pid() and AXUI element lookups in the window to avoid a system call on every move.

karinushka:
@Fjx-dylanZ - can you check out this latest commit on the `testing` branch: https://github.com/karinushka/paneru/commit/04838b316aa769602fa736140ea7038814324138
Apparently it is sufficient to just disable the enhanced user interface once? (doing it once in https://github.com/karinushka/paneru/commit/8c49832eed03ba1b5ade8a6efd8eda756c3b2fff)

If this still works for your usecase, then it's a more efficient approach.

(I upstreamed this change among others, from @mkdir700's excellent fork https://github.com/karinushka/paneru/compare/main...mkdir700:paneru:main)

Fjx-dylanZ:
Tested https://github.com/karinushka/paneru/commit/04838b316aa769602fa736140ea7038814324138 and unfortunately the lag returns for me...

I suspect it depends on whether an app actively re-asserts `AXEnhancedUserInterface` after we disable it, if so, the one-shot approach in https://github.com/karinushka/paneru/commit/04838b316aa769602fa736140ea7038814324138 won't hold. For users whose apps don't re-assert it, the one-shot method should be sufficient.

Since it's app-dependent, could we expose a config flag so users can opt into either method (per-call vs. once)? Happy to PR if that sounds reasonable.

karinushka:
It will be confusing for users to see some apps lagging - probably better to leave it in. It didn't seem to affect the performance so doesn't matter - was just to make stuff simpler.
Thanks for testing! I'll rollback the change.

PR #187, opened 2026-04-15
fix: Undim windows on quit

ToroidalFox:
Should be very straightforward.
Although as I write this I noticed that changing config should also update dimming. Which is not as big of an issue as not undimming on quit, so maybe later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2554,6 +2554,7 @@ dependencies = [
  "bitflags 2.11.0",
  "chrono",
  "clap",
+ "ctrlc",
  "derive_more",
  "embed_plist",
  "launchctl",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,4 @@ stdext = "0.3.3"
 toml = "1.1.2"
 tracing = "0.1.44"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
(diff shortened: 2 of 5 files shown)


karinushka:
Is the existing `cleanup_on_exit()` not working for you?
https://github.com/karinushka/paneru/blob/a66d906d5426c65d35dc70f7ecfb844b1e8b28f6/src/ecs/systems.rs#L1193-L1206

Had dimming disabled for a couple of weeks, but just tried enabling it and exited - and all the windows are undimmed.

Yeah, disabling dimming in the config is a bit dodgy 😺 It leaves the windows dimmed, need a restart.

ToroidalFox:
It works. ~Now I don't remember/know what made me so convinced that undimming doesn't work lol.~ And just as I write that, found it. it doesn't behave correctly with SIGTERM, likely from `paneru stop`.

Welp, that was pain to iterate and debug. Please check new commits.

karinushka:
Oh, so `paneru stop`just `SIGTERM`'s the service? Maybe it should be adjusted to send a proper `quit` command down the socket.

ToroidalFox:
yep. `launchctl::Service::stop` does SIGTERM. I think it should stay as handling SIGINT and SIGTERM as this also fixes foreground execution behavior.

PR #198, opened 2026-04-27
fix: don't tile or reposition windows during init

nikosdouvlis:
On a multi-monitor setup, paneru launching with windows on inactive displays drags every window onto the active display within ~120ms. Fires on every cold start.

`apply_window_properties` appends every new window to `active_display.active_strip()` regardless of physical display. `finish_setup` later reassigns each window to its real strip from `windows_in_workspace`. The problem is timing: the layout writers run within the same tick using the pre-`finish_setup` strips, cache active-display coordinates into each window's `Position`, and the next `commit_window_position` pushes those to macOS.

Gate the layout writers and the macOS commit systems on `!Initializing`, and order the layout group `.after(finish_setup)` so it only sees corrected strip membership. After init the layout chain runs once with the right inputs and commits in a single pass.

Three other init-time paths bypass the ECS layer and call `window.reposition` / `window.resize` directly against `active_display.bounds()`. Same root issue. Gate them too:

- `apply_window_defaults` (floating + `grid_ratios`, configured `width`)
- `window_unmanaged_trigger` (clamping a new floating window into the active viewport)
- `window_managed_trigger` (reachable via `WindowDeminimized` / `ApplicationVisible`)

Strip removal in `window_unmanaged_trigger` still runs during init so floating windows aren't double-tracked.

Test plan:
- Manual: lid closed, LC49 (active) + external monitor, windows on both. Restart paneru. Each monitor keeps its windows.
- `cargo build`, `cargo clippy --all-targets`, `cargo test` clean (61 tests pass).

### DIFF
--- a/src/ecs/triggers.rs
+++ b/src/ecs/triggers.rs
@@ -524,36 +524,41 @@ pub(super) fn window_unmanaged_trigger(
 
     let properties = WindowProperties::new(app, window, config.config());
 
-    if let Some((rx, ry, rw, rh)) = properties.grid_ratios() {
-        let x = (f64::from(display_bounds.width()) * rx) as i32;
-        let y = (f64::from(display_bounds.height()) * ry) as i32;
-        let w = (f64::from(display_bounds.width()) * rw) as i32;
-        let h = (f64::from(display_bounds.height()) * rh) as i32;
-        reposition_entity(entity, Origin::new(x, y), &mut commands);
-        resize_entity(entity, Size::new(w, h), &mut commands);
-    } else {
-        let max_width = display_bounds.width() * UNMANAGED_MAX_SCREEN_RATIO_NUM
-            / UNMANAGED_MAX_SCREEN_RATIO_DEN;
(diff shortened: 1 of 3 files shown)


src/ecs/triggers.rs
```diff
@@ -859,6 +869,7 @@ fn apply_window_defaults(
     active_display: &mut ActiveDisplayMut,
     properties: &[WindowParams],
     config: &Config,
+    initializing: bool,
```
karinushka: nit: If you pass `&Configuration` here, then you won't need the boolean - you can directly use the `config.initializing()` like in the other place.

karinushka (approved):

src/ecs/triggers.rs
nikosdouvlis: Done, swapped to &Configuration, thank you!

nikosdouvlis: This is now good from my end, feel free to merge at your leisure 🙏🏻

PR #204, opened 2026-05-04
feat(api): add JSON state and absolute workspaces

kiwnix:
- Add `paneru query` and `paneru subscribe` commands for structured JSON state over the Unix socket
- Support absolute virtual workspace switching via `window virtualnum` and related commands
- Preserve empty virtual workspace rows to maintain stable numbered slots

Note: Im not an expert in rust, I used codex for helping me code this.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -8,6 +8,8 @@ use bevy::math::IRect;
 use tracing::{Level, instrument};
 use tracing::{debug, info};
 
+mod query;
+
 use crate::config::Config;
 use crate::ecs::layout::{Column, LayoutStrip, StackItem};
 use crate::ecs::params::{ActiveDisplay, ActiveDisplayMut, Windows};
@@ -112,6 +114,7 @@ pub enum Command {
 }
 
 pub fn register_commands(app: &mut bevy::app::App) {
+    query::register_query_commands(app);
     app.add_systems(
         PreUpdate,
         (
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -84,8 +84,12 @@ pub enum Operation {
     Snap,
     /// Cyclically selects the virtual strip for the current workspace.
     Virtual(Direction),
+    /// Selects a virtual strip by its zero-based index for the current workspace.
+    VirtualNumber(u32),
     /// Moves the focused window to the virtual strip.
     VirtualMove(Direction, MoveFocus),
+    /// Moves the focused window to a virtual strip by its zero-based index.
+    VirtualMoveNumber(u32, MoveFocus),
 }
 
 /// Defines operations that can be performed on the mouse.
--- a/src/ecs/workspace.rs
+++ b/src/ecs/workspace.rs
@@ -43,7 +43,6 @@ impl Plugin for WorkspaceEventsPlugin {
             Update,
             (
                 show_active_workspace,
-                cleanup_virtual_workspaces,
                 handle_virtual_window_moves,
                 detect_moved_windows.run_if(not(resource_exists::<Initializing>)),
                 refresh_workspace_window_sizes.run_if(on_timer(Duration::from_millis(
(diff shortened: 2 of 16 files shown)


src/ecs/workspace.rs
```diff
@@ -560,16 +560,15 @@ fn cleanup_selected_space_marker(
 #[allow(clippy::needless_pass_by_value)]
 fn cleanup_virtual_workspaces(
```
karinushka: This cleanup_virtual_workspaces() no longer does any work. So you can just delete it.

src/ecs/workspace.rs
```diff
@@ -727,12 +715,14 @@ fn handle_virtual_window_moves(
 fn switch_virtual_workspace_bind(
     mut messages: MessageReader<Event>,
     active_display: ActiveDisplay,
+    active_display_entity: Single<Entity, With<ActiveDisplayMarker>>,
```
karinushka: Add a helper to the ActiveDisplay system param:
```rust
pub fn entity(&self) -> Entity {
   self.display.1
}
```
Then you don't need this `active_display_entity` argument and can just replace the `*active_display_entity` with `active_display.entity()`.

src/ecs/workspace.rs
```diff
@@ -748,9 +738,35 @@ fn switch_virtual_workspace_bind(
...
+                    .spawn((
+                        strip,
+                        Position(active_display.bounds().min),
+                        ChildOf(*active_display_entity),
+                    ))
+                    .id();
+                if let Ok(mut entity_commands) = commands.get_entity(entity) {
+                    entity_commands
+                        .try_insert(SelectedVirtualMarker)
+                        .try_insert(ActiveWorkspaceMarker);
```
karinushka: These inserts can be folded up into the creation of the entity above:
```rust
commands.spawn((
  strip,
  Position(active_display.bounds().min),
  ChildOf(*active_display_entity),
  SelectedVirtualMarker,
  ActiveWorkspaceMarker,
))
```

src/commands.rs
```diff
@@ -854,6 +926,207 @@ fn snap_window(
...
+
+    if outgoing.is_empty() {
+        return;
+    }
+
+    let mut payload = String::new();
+    for event in outgoing {
+        payload.push_str(&event.to_string());
+        payload.push('\n');
+    }
```
karinushka:
```rust
    let payload = outgoing
        .into_iter()
        .map(|event| event.to_string())
        .collect::<Vec<_>>()
        .join("\n");
```
General optimization 😺
It is more efficient to create a list of things and then join them into a string, than repeatedly append stuff to a string. That's 100% the case in Python, but probably also is so in Rust.

src/commands.rs
```diff
@@ -854,6 +926,207 @@ fn snap_window(
     );
 }
 
+#[allow(clippy::needless_pass_by_value, clippy::type_complexity)]
```
karinushka: nit: The types are fine - no `clippy::type_complexity` there 😺
Try removing those clippy overrides first, do a `cargo clippy` and only add where they are needed.

karinushka (commented):
Looks good overall - added some small nit comments.

There seems to be a race-condition in the tests somewhere. Running them a bunch of times in a row causes 30% attempts or so to fail with an abort:
```bash
$ for att in `seq 1 30` ; do cargo test ; done

...
test tests::interaction::test_scrolling_stop ... ok
test tests::display::test_multi_display_no_height_crosstalk ... ok
test tests::interaction::test_dont_focus ... ok
error: test failed, to rerun pass `--bin paneru`

Caused by:
  process didn't exit successfully: `/Users/karinushka/rust/paneru/target/debug/deps/paneru-c432c1d34259ab9c` (signal: 6, SIGABRT: process abort signal)
```

Also, I am wondering whether these Query related things should go into a stand-alone `query.rs` file? Otherwise `commands.rs` and `state.rs` become too large. What do you think?

Edit: also maybe split this into two commits for clarity - one which adds the virtual workspace numerics and the other one which adds the socket query feature.

kiwnix:
Hi, Thanks for the review. I addressed the feedback and cleaned up the branch.

I made the followin changes:
- Rebased the branch on top of `testing`.
- Split the work into two commits:
  - `feat: Add numbered virtual workspace commands`
  - `feat: Add JSON state query and subscribe commands`
- Moved the query/subscription implementation into `src/commands/query.rs`.
- Removed the no-op virtual workspace cleanup system.
- Simplified the virtual workspace spawn path and reused `ActiveDisplay::entity()`.
- I think i fixed the config parser test race by injecting deterministic virtual key maps in tests instead of calling the live macOS keyboard layout API.

karinushka:
Thank you for helping out!
Do you have an example configuration of Paneru with Sketchybar? If so maybe drop it into a discussion thread as an example for other users.

kiwnix:
Hi @karinushka , i have an example Sketchybar configuration extracted from my nix home-manager config in https://github.com/kiwnix/paneru/tree/feat/example-sketchybar/assets/example-sketchybar , but i don't know if its the correct place (having it in a gist "with folders" is a bit unclear). Its WIP and maybe not the most "simple one".

PR #209, opened 2026-05-05
fix: accept device-independent modifier flags from synthesized keystrokes

### LINKED ISSUE #210: Keybindings ignored when triggered by synthesized keystrokes (e.g. Better Mouse button → ⌥→)
## Summary

Keybindings configured with modifiers (e.g. `alt - rightarrow`) do not fire when the keystroke is synthesized by external utilities like [Better Mouse](https://better-mouse.com/) or some Karabiner-Elements complex modifications, even though the same keystroke pressed on a real keyboard works correctly.

## Repro

1. In Better Mouse, map a mouse button (e.g. Button 4) to `⌥→`.
2. In `paneru.toml`, ensure `window_focus_east = "alt - rightarrow"` (default).
3. Press `⌥→` on the keyboard → focus moves east. ✅
4. Click the mapped mouse button → nothing happens. ❌

## Root cause

`get_modifiers()` in `src/platform/input.rs` reads only the device-dependent left/right modifier bits from `CGEventFlags` (`0x0000_0020` for LAlt, `0x0000_0040` for RAlt, etc.). Synthesized keystrokes from these utilities set only the **device-independent** `kCGEventFlagMask*` bits (`0x0008_0000` for Alt, `0x0010_0000` for Cmd, etc.) without the L/R bits. The mask is parsed as `Modifiers::empty()`, the binding lookup returns `None`, and the action never fires.

There was even an explicit test (`device_independent_flags_ignored`) confirming this was the established behavior, but the implications for synthesized input weren't covered.

## Fix

PR #209 adds a fallback: when no L/R bit is set in a modifier group, the device-independent flag is used and both sides are marked (`LALT|RALT`). This matches generic `alt - …` bindings and naturally fails `lalt`/`ralt`-specific ones, which a sideless synthetic event can't satisfy anyway. Hardware key events still take precedence because they set the L/R bits, which the loop processes first.

## Affected tools

- Better Mouse (mouse button → key combo mappings)
- Some Karabiner-Elements setups, depending on how the manipulator is configured
- Any AppleScript / `CGEventCreateKeyboardEvent` based remapper that doesn't manually OR in the device-dependent flags

bensleveritt:
Closes #210.

## Summary

- Tools that synthesize keystrokes — Better Mouse mapping a mouse button to `⌥→`, some Karabiner setups, etc. — set only the device-independent `kCGEventFlagMask*` bits without the device-dependent left/right modifier bits paneru's event tap was reading. The modifier mask arrived as `Modifiers::empty()`, the keybinding lookup never matched, and the configured action never fired.
- `get_modifiers()` now falls back to the device-independent flags when the L/R bits are absent. The fallback marks both sides (e.g. `Modifiers::ALT` = `LALT|RALT`) since the flag doesn't tell us which side is held; this matches generic `alt - …` bindings and naturally fails `lalt`/`ralt`-specific ones, which a sideless synthetic event can't satisfy anyway.
- Real hardware key events still take precedence: the L/R loop runs first, and the fallback only kicks in when no L/R bit is set in that group. Added a regression test (`device_dependent_takes_precedence_over_independent`) covering this.

## Why the previous behavior existed

The original `device_independent_flags_ignored` test was deliberate — `CGEventFlags` carries both device-dependent and device-independent bits for hardware events, and double-counting them would have inflated the mask. The new code preserves that invariant (only falls back when no device-dependent bit is set) while handling synthesized events whose only signal is the device-independent flag.

## Test plan

- [x] `cargo test platform::input::tests` — 10/10 pass, including 4 new tests covering Alt/Shift/Cmd/Ctrl device-independent fallback, combined fallback, and device-dependent precedence.
- [x] `cargo build` — clean.
- [ ] Manual: bind a Better Mouse button to `⌥→` (or any combo matching a paneru keybinding) and confirm the action now fires.

### DIFF
--- a/src/platform/input.rs
+++ b/src/platform/input.rs
@@ -423,10 +423,27 @@ fn get_modifiers(eventflags: CGEventFlags) -> Modifiers {
         (Modifiers::LCTRL, 0x0000_0001),
         (Modifiers::RCTRL, 0x0000_2000),
     ];
+    // Device-independent fallbacks (kCGEventFlagMask*). Used by software
+    // that synthesizes keystrokes without setting the device-dependent
+    // L/R bits (e.g. Better Mouse mapping a button to ⌥→). The flag
+    // does not say which side, so we mark both — this matches generic
+    // `alt - …` bindings and naturally fails `lalt`/`ralt`-specific ones,
+    // which a sideless synthetic event can't satisfy anyway.
+    const GENERIC_MASKS: [(Modifiers, u64); 4] = [
+        (Modifiers::ALT, 0x0008_0000),
+        (Modifiers::SHIFT, 0x0002_0000),
+        (Modifiers::CMD, 0x0010_0000),
+        (Modifiers::CTRL, 0x0004_0000),
+    ];
     let mut mask = Modifiers::empty();
     for (modifier, m) in MODIFIER_MASKS {
         if eventflags.0 & m != 0 {
             mask |= modifier;
+        }
+    }
+    for (group, m) in GENERIC_MASKS {
+        if eventflags.0 & m != 0 && !mask.intersects(group) {
+            mask |= group;
         }
     }
     mask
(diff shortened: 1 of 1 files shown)


bensleveritt:
Closing — the fix targets device-independent modifier flags, but the actual bug we set out to address (Better Mouse mouse-button → ⌥→) doesn't reach paneru's HID-level event tap at all (Better Mouse posts at the session level). Reopening if a confirmed real-world repro for the device-independent flag case turns up.

PR #211, opened 2026-05-06
feat: improve horizontal_mouse_warp landing and add vertical offset

bensleveritt:
## Summary

Polishes the existing `horizontal_mouse_warp` feature so that the cursor flow across vertically-arranged displays feels continuous, matches macOS's native side-by-side behavior, and doesn't suffer the post-warp "standing start" caused by `CGWarpMouseCursorPosition`'s default event-suppression interval. Also adds a config option to compensate for physical desk arrangements that don't match the macOS arrangement.

Tested on a setup with a built-in laptop display (1512×982) above a 1080×1920 portrait monitor in macOS arrangement.

## Changes

1. **Land at the opposite edge of the target** instead of its center, preserving cursor flow: leaving the right edge of one display reappears just inside the left edge of the next, and vice versa. A small inset (6 px) prevents immediate re-warping.

2. **Preserve relative Y** from the source display's top so vertical motion across the warp feels continuous — matches macOS's behavior for side-by-side displays.

3. **Carry over horizontal velocity** by extrapolating the cursor's pre-warp motion ~30 ms into the target (capped at 80 px). A fast flick lands further in; a slow drift lands at the inset. Stale samples (>80 ms) are discarded.

4. **Drop the local-event suppression interval to zero** in `warp_mouse` and re-call `CGAssociateMouseAndMouseCursorPosition(true)` so HID mouse events resume immediately after the warp. The default 250 ms interval was dropping the user's continuing physical motion, which felt like the cursor was "stuck" at the warp landing position.

5. **Skip the warp when no equivalent Y exists on the target.** For displays of unequal height (e.g. tall portrait beside a shorter laptop), the cursor's relative Y position may map off the top or bottom of the target. The previous code would clamp; this commit blocks the warp entirely so the user can only cross at Y positions where both displays exist — matching macOS's native side-by-side behavior.

6. **New `horizontal_mouse_warp_offset` config option** (Integer, px). Signed by warp direction: positive values land the cursor lower when warping to a display *below* (in macOS arrangement) and higher when warping to one *above*. Compensates for physical desk arrangement differing from the macOS arrangement.

## Tradeoffs

- **Velocity tracking uses a `Local<WarpVelocityState>`.** It's purely local to the observer and reset across stale events — no cross-system coupling.
- **The 6 px landing inset is a constant.** Could be configurable, but most users won't need to tune it.
- **The 30 ms / 80 px carry parameters** were arrived at empirically from in-hand testing. Open to adjustments.
- **Skip-warp on out-of-range Y** is a behavioral change from "clamp to edge". Users who relied on the clamp will see warps blocked instead. The new behavior matches macOS's native side-by-side semantics, which is what `horizontal_mouse_warp` is simulating in the first place.

## Test plan

- [x] `cargo build --release` clean.
- [x] `cargo test` — 65/66 pass; the one failure (`config::test_config_defaults`) is preexisting on `main` and unrelated.
- [x] Manual: laptop right edge → portrait, cursor lands at portrait's left edge with continuous Y, physical mouse motion takes over without pause.
- [x] Manual: portrait left edge → laptop, same.
- [x] Manual: portrait bottom region (Y > laptop_height − offset) → warp blocked, cursor stays on portrait.
- [x] Manual: tuned `horizontal_mouse_warp_offset = 250` for a portrait monitor sitting physically below the laptop, both directions feel right.
- [ ] Cross-platform: only tested on aarch64-darwin / macOS 15.

## CONFIGURATION.md

Updated the existing `horizontal_mouse_warp` row and added a new row for `horizontal_mouse_warp_offset`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


karinushka:
Thank you for your contribution!

I am sure @rami3l will love this PR as the warp mouse user 😺

PR #223, opened 2026-05-15
feat: Restore window sessions on startup

kiwnix:
Save managed window state to the user state directory and restore it during a
bounded startup phase. Restore now rebuilds saved layouts, virtual workspace
rows, active virtual selection, and display associations for matched windows.

Add restore configuration for enabling startup restore, setting the startup
grace period, and choosing the missing-window policy. Missing saved windows are
ignored by default and restored layouts are compacted.

Matched startup windows use saved session placement before static window rules,
while unmatched windows and post-restore windows keep normal config behavior.


AndreiTich:
This is awesome, thank you for this! It may be good to also update some context around `ARCHITECTURE.md` as well as the readme as this is a pretty large addition.

kiwnix:
I rebased and added some docs. As my other PR, i'm not an expert in rust or macos software development and i have used codex.

karinushka:
> I rebased and added some docs. As my other PR, i'm not an expert in rust or macos software development and i have used codex.

Thank you for contributing! Looks good!
This is yet another advantage of using Bevy - the decoupled nature of ECS makes it easy for an LLM to insert additional functionality. You just need to guide it towards these principles, otherwise it may stray in the direction of normal Rust and start inserting Arc\<Mutex\>'es everywhere. 😺

PR #225, opened 2026-05-15
fix: Full-width mode fights layout pipeline, breaking stacked windows

AndreiTich:
The layout pipeline (layout_strip_changed, position_layout_windows) had no awareness of FullWidthMarker and would overwrite the window's Bounds and Position every frame, causing jitter and half-width rendering.

Additionally, full-width mode only expanded width but kept the stacked height, and exiting full-width would restack the window — causing confusing state transitions.

### Changes:
- Skip FullWidthMarker windows in layout_strip_changed and position_layout_windows so the animation system has sole control
- Expand to full viewport width AND height when entering full-width
- Store height_ratio in FullWidthMarker (replaces was_stacked)
- On exit, set Bounds directly instead of animating to avoid layout pipeline fight on re-entry
- Exit full-width automatically when stacking a full-width window
- Extract exit_full_width helper to deduplicate the restore logic

### Before:

https://github.com/user-attachments/assets/9d0954b0-2c9c-4c59-b9df-0f2ca552f5a6

### After:

https://github.com/user-attachments/assets/34c1aeb7-4803-47e5-932d-90464543834c

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -12,9 +12,9 @@ use crate::config::Config;
 use crate::ecs::layout::{Column, LayoutStrip, StackItem};
 use crate::ecs::params::{ActiveDisplay, ActiveDisplayMut, Windows};
 use crate::ecs::{
-    ActiveDisplayMarker, ActiveWorkspaceMarker, FocusFollowsMouse, FocusedMarker, FullWidthMarker,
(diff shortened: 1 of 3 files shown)


karinushka:
Cool stuff! Thanks for fixing it - I ran into it at some point and had it on my internal TODO list, but you beat me to it!
😺

PR #232, opened 2026-05-18
Horizontal window swap animation

soraxas:
# Current behaviour

https://github.com/user-attachments/assets/201a2a78-16b8-45d4-86cf-ab4d8e84ea07

# After PR

https://github.com/user-attachments/assets/80201944-c74a-4862-a85f-0561b43fa727

### DIFF
--- a/src/ecs/layout.rs
+++ b/src/ecs/layout.rs
@@ -1,5 +1,5 @@
 use bevy::app::{App, Plugin, Update};
-use bevy::ecs::change_detection::DetectChangesMut as _;
+use bevy::ecs::change_detection::{DetectChanges as _, DetectChangesMut as _};
 use bevy::ecs::component::Component;
 use bevy::ecs::entity::Entity;
 use bevy::ecs::hierarchy::ChildOf;
@@ -7,6 +7,7 @@ use bevy::ecs::query::{Changed, Has, Or, With, Without};
 use bevy::ecs::schedule::IntoScheduleConfigs as _;
 use bevy::ecs::schedule::common_conditions::{not, resource_exists};
 use bevy::ecs::system::{ParallelCommands, Populated, Query, Res};
+use bevy::ecs::world::Ref;
 use bevy::math::IRect;
 use std::collections::VecDeque;
 use stdext::function_name;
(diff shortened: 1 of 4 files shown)


src/ecs/layout.rs
```diff
@@ -859,6 +861,57 @@ fn reshuffle_layout_strip(
...
+#[instrument(level = Level::DEBUG, skip_all)]
+fn ensure_visible_in_strip(
+    markers: Populated<(Entity, &LayoutPosition), With<EnsureVisibleMarker>>,
+    strips: Query<(&LayoutStrip, Entity, &Position, &ChildOf)>,
+    displays: Query<(&Display, Option<&DockPosition>)>,
+    windows: Windows,
+    config: Res<Config>,
+    commands: ParallelCommands,
+) {
+    markers.par_iter().for_each(|(entity, layout_position)| {
```
karinushka, line 880: I was using these parallel iterators a lot, but then I realized that thread pool setup takes some overhead - so the official documentation recommends only to use them if you have thousands of items to process.
(I have a commit in `testing` branch which removes all these parallel iterators: https://github.com/karinushka/paneru/commit/c9e1011d10bce6307141030b7039167f19a4a25f)

So better use a normal for loop.
Note: if using a for loop, all the returns need to be changed to `continue` to avoid subtle bugs 😺

src/ecs/layout.rs
```diff
@@ -904,11 +967,19 @@ fn position_layout_windows(
...
             else {
                 return;
             };
+            let strip_position = strip_pos.0;
+            // Workspace switch: show_active_workspace directly mutates the
+            // strip's Position to slide the strip on/off-screen — no
+            // RepositionMarker, no Scrolling. We snap windows in lockstep so
+            // they don't animate the long jump from off-screen storage into
+            // view. (See CLAUDE.md "Reverting the workspace-switch snap" if
+            // you want to make this animate later.)
```
karinushka, line 981: Add a TODO here instead of Claude 😺

karinushka:
Thanks for the PR! The animated swap looks very nice.
Can you rebase this onto `testing` branch to reduce chance of wider breakage.

karinushka:
I rebased this on top of the `testing` branch with you as the author: https://github.com/karinushka/paneru/commit/32c83836ed1d235075342cc2945472ea641fd44d

Fetch it and check whether it works as you intended.

karinushka:
> I rebased this on top of the `testing` branch with you as the author: [32c8383](https://github.com/karinushka/paneru/commit/32c83836ed1d235075342cc2945472ea641fd44d)
> 
> Fetch it and check whether it works as you intended.

There was also a small bug which didn't detect the moving workspaces correctly. Fixed it in a follow up: https://github.com/karinushka/paneru/commit/c38f172483c21ff6d6093d709d7774593e876101

soraxas:
Yes it seems to work for me! But the animation when running `window_virtualmove_north/south` is a bit dis-orienting (when windows animate from hidden state to their strip position) so my local patch had detect and disable them. What do you think?

karinushka:
> is a bit dis-orienting (when windows animate from hidden state to their strip position) 

Probably okay, since it gives a visual indication that a window is being moved. 🤷

soraxas:
Patch was merged and changes here no longer valid

PR #234, opened 2026-05-18
perf: Improve various performance points

kiwnix:
Various performance improvements (paneru is more responsive with this changes). I splited it in invididual commits to allow better review and modular changes/acceptance if needed.

- Reduces event loop latency by draining ready events before the Cocoa pump, coalescing `MouseMoved` events, and capping waits while visible work or low-power mode is active.
- Avoids expensive hot-path work: fewer AX reads during spawn logging, nonblocking subscriber sockets, hard-match caching for session restore, and fallback metadata reads only when needed.
- Optimizes layout and window commits by caching `LayoutStrip` membership/context, using `HashSet` lookups for unresolved windows, and combining position/size updates into a single `set_frame` commit.
- Cleans up lifecycle resources by unregistering one-shot timeout systems during early cleanup and avoiding retention of unused AX observer contexts.

Im not an expert in rust or macos-development, i writed this changes with the help of codex

### DIFF
--- a/src/ecs.rs
+++ b/src/ecs.rs
@@ -73,7 +73,11 @@ pub fn register_systems(app: &mut bevy::app::App) {
     );
     app.add_systems(
         PreUpdate,
-        (systems::window_creation_event, systems::pump_events),
+        (
+            systems::window_creation_event,
+            systems::detect_tabbed_windows,
+            systems::pump_events,
+        ),
     );
     app.add_systems(
         Update,
--- a/src/ecs.rs
+++ b/src/ecs.rs
@@ -157,6 +157,8 @@ pub fn register_triggers(app: &mut bevy::app::App) {
         .add_observer(triggers::send_message_trigger)
         .add_observer(triggers::window_removal_trigger)
         .add_observer(systems::cleanup_timeout)
+        .add_observer(systems::release_frame_commit_after_reposition)
+        .add_observer(systems::release_frame_commit_after_resize)
         .add_observer(restore::restore_window_state);
 }
 
--- a/src/ecs.rs
+++ b/src/ecs.rs
@@ -156,7 +156,7 @@ pub fn register_triggers(app: &mut bevy::app::App) {
         .add_observer(triggers::locate_dock_trigger)
         .add_observer(triggers::send_message_trigger)
         .add_observer(triggers::window_removal_trigger)
-        .add_observer(triggers::cleanup_timeout_trigger)
+        .add_observer(systems::cleanup_timeout)
         .add_observer(restore::restore_window_state);
 }
 
@@ -463,6 +463,18 @@ pub fn flash_message(message: String, duration: f32, commands: &mut Commands) {
     commands.spawn((timeout, FlashMessage(message)));
 }
 
+pub fn remove_timeout(entity: Entity, commands: &mut Commands) {
+    if let Ok(mut entity_commands) = commands.get_entity(entity) {
+        entity_commands.try_remove::<Timeout>();
+    }
+}
+
+pub fn despawn_timeout_entity(entity: Entity, commands: &mut Commands) {
+    if let Ok(mut entity_commands) = commands.get_entity(entity) {
+        entity_commands.try_despawn();
+    }
+}
+
 pub fn setup_bevy_app(sender: EventSender, receiver: Receiver<Event>) -> Result<BevyApp> {
     let window_manager: Box<dyn WindowManagerApi> = Box::new(WindowManagerOS::new(sender.clone()));
     let watcher = window_manager.setup_config_watcher(CONFIGURATION_FILE.as_path())?;
@@ -574,5 +586,64 @@ impl WindowProperties {
             .iter()
             .find_map(|props| props.horizontal_padding)
             .unwrap_or(0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use bevy::app::App;
+    use bevy::ecs::resource::Resource;
+    use bevy::ecs::system::ResMut;
+    use bevy::time::TimerMode;
+
+    #[derive(Default, Resource)]
+    struct TimeoutTestRuns(u32);
+
+    fn count_timeout_run(mut runs: ResMut<TimeoutTestRuns>) {
+        runs.0 += 1;
+    }
+
+    fn registered_timeout(app: &mut App) -> (Entity, SystemId) {
+        let system_id = app.world_mut().register_system(count_timeout_run);
+        let timer = Timer::from_seconds(1.0, TimerMode::Once);
+        let entity = app
+            .world_mut()
+            .spawn(Timeout {
+                timer,
+                system_id: Some(system_id),
+            })
+            .id();
+        (entity, system_id)
+    }
+
+    fn timeout_cleanup_app() -> App {
+        let mut app = App::new();
+        app.init_resource::<TimeoutTestRuns>()
+            .add_observer(systems::cleanup_timeout);
+        app
+    }
+
+    #[test]
+    fn timeout_unregisters_system_when_removed() {
+        let mut app = timeout_cleanup_app();
+        let (entity, system_id) = registered_timeout(&mut app);
+
+        app.world_mut().entity_mut(entity).remove::<Timeout>();
+        app.world_mut().flush();
+
+        assert!(app.world_mut().run_system(system_id).is_err());
+    }
+
+    #[test]
+    fn timeout_unregisters_system_when_despawned() {
+        let mut app = timeout_cleanup_app();
+        let (entity, system_id) = registered_timeout(&mut app);
+
+        app.world_mut().entity_mut(entity).despawn();
+        app.world_mut().flush();
+
+        assert!(app.world_mut().run_system(system_id).is_err());
     }
 }
--- a/src/ecs/systems.rs
+++ b/src/ecs/systems.rs
@@ -899,16 +899,53 @@ pub(super) fn verify_window_position(
 pub(super) fn commit_window_frames(
     active_display: ActiveDisplay,
     mut framed_windows: Populated<
-        (&mut Window, Ref<Position>, Ref<Bounds>, &mut WidthRatio),
+        (
+            &mut Window,
+            Ref<Position>,
+            Ref<Bounds>,
+            &mut WidthRatio,
+            Option<&RepositionMarker>,
+            Option<&ResizeMarker>,
+        ),
         Or<(Changed<Position>, Changed<Bounds>)>,
     >,
 ) {
     let display_bounds = active_display.bounds();
-    for (mut window, position, bounds, mut width_ratio) in &mut framed_windows {
+    for (mut window, position, bounds, mut width_ratio, reposition, resize) in &mut framed_windows {
         if bounds.is_changed() {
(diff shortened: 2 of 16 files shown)


karinushka (commented):
Thanks a lot for looking into performance!
Overall looks very good - I've left a few comments here and there, let's discuss whether some changes feel somewhat placebo like 😺

If you want you can pull the commits which are ready (without any discussion points) into a separate PR and we can push it faster.

src/ecs/systems.rs
```diff
@@ -662,6 +680,19 @@ pub(super) fn pump_events(
...
+        return;
+    }
+    if !ready.events.is_empty() {
+        for event in ready.events {
+            messages.write(event);
+        }
+        *timeout = LOOP_TIMEOUT_STEP;
+        return;
+    }
+
```
karinushka: I am not sure this gives any performance benefit - it pre-drains the event queue for any potential pre-existing events. But this should be rare, because macOS side events happen when we pump the cocoa loop below and they bubble up from the platform.rs side.
So logically the flow would be - pump the loop and then reap any generated events afterwards.

src/manager/windows.rs
```diff
@@ -526,7 +526,7 @@ impl WindowApi for WindowOS {
 
             // Artificially delay the activation by 1ms. This is necessary because some
             // applications appear to be confused if both of the events appear instantaneously.
-            thread::sleep(Duration::from_millis(20));
+            thread::sleep(Duration::from_millis(1));
```
karinushka: This sleep is a cargo-cult from Yabai where the sleep is 40ms: https://github.com/asmvik/yabai/blob/2d743f4bea98b650ae7dec9f459db600323ca29d/src/window_manager.c#L1306-L1313

Maybe in the modern macOS versions this is no longer required, but risky to remove it and break older version. 🤷

src/ecs/systems.rs
```diff
@@ -965,9 +1010,9 @@ pub(super) fn update_overlays(
 pub(super) fn commit_window_position(
     mut moved_windows: Populated<(&mut Window, &Position), Changed<Position>>,
 ) {
-    moved_windows
-        .par_iter_mut()
-        .for_each(|(mut window, position)| window.reposition(position.0));
+    for (mut window, position) in &mut moved_windows {
+        window.reposition(position.0);
+    }
```
karinushka: This one is dubious - setting up a parallel iterator costs some overhead, so it should not be done if the number of items is lower than a thousand or so...
But in this case it calls a slow macOS function - so maybe doing it in parallel is worth it? What does the LLM suggest?

src/ecs/systems.rs
```diff
@@ -694,23 +726,36 @@ pub(super) fn pump_events(
...
+    let mut received_events = Vec::new();
+    let mut pending_mouse = None;
     loop {
         // Repeatedly drain the events until timeout.
         match incoming_events.recv_timeout(Duration::from_millis(1)) {
             Ok(Event::Exit) | Err(RecvTimeoutError::Disconnected) => {
+                flush_pending_mouse(&mut received_events, &mut pending_mouse);
+                for event in received_events {
+                    messages.write(event);
+                }
```
karinushka: If the app is exiting, we probably don't care about delivering a few last mouse movement events 😺

src/ecs/systems.rs
```diff
@@ -694,23 +726,36 @@ pub(super) fn pump_events(
...
                 exit.write(AppExit::Success);
                 break;
             }
             Ok(event) => {
-                messages.write(event);
+                push_coalesced_event(&mut received_events, &mut pending_mouse, event);
```
karinushka: This function can be inlined as a single `if` for simplicity, making it easier to follow what's going on. E.g.:
```rust
match event {
Mouse(mouse) => pending_mouse = mouse
_ => received_events.push(event)
}
```

src/ecs/systems.rs
```diff
@@ -694,23 +726,36 @@ pub(super) fn pump_events(
...
             }
             Err(RecvTimeoutError::Timeout) => {
-                let timeout_limit = if low_power_mode.is_some_and(|low_power| low_power.0) {
-                    LOOP_MAX_TIMEOUT_LOWPOWER_MS
-                } else {
-                    LOOP_MAX_TIMEOUT_MS
-                };
+                flush_pending_mouse(&mut received_events, &mut pending_mouse);
+                for event in received_events {
+                    messages.write(event);
```
karinushka: There is a `write_batch` variant which takes an iterator - should be more efficient.
So basically:
- Is there something in the `pending_mouse`? Append it to the vector.
- Batch write the whole vector.

src/manager/windows.rs
```diff
@@ -297,6 +301,62 @@ impl WindowOS {
         event_bytes[0x08] = 0x02;
         unsafe { SLPSPostEventRecordTo(psn, event_bytes.as_ptr().cast()) };
     }
+
+    fn set_position_attribute(&mut self, origin: Origin) -> bool {
```
karinushka, line 338: This and the size helper below should be in the `impl WindowOS {...}` section, because they are not generic enough for the interface (trait). They are calling macOS system functions.

src/manager/windows.rs
```diff
@@ -561,10 +634,6 @@ impl WindowApi for WindowOS {
                 SLPSPostEventRecordTo(&focused_psn, event_bytes.as_ptr().cast());
             }
 
-            // Artificially delay the activation by 1ms. This is necessary because some
-            // applications appear to be confused if both of the events appear instantaneously.
-            thread::sleep(Duration::from_millis(1));
-
```
karinushka: Ooof.. here it is completely removed.

We can potentially have it as a configuration toggle - by default it's disabled, but if people notice issues with focus-follows-mouse, they can toggle this "legacy tweak".

src/ecs.rs
```diff
@@ -444,6 +444,24 @@ pub fn flash_message(message: String, duration: f32, commands: &mut Commands) {
     commands.spawn((timeout, FlashMessage(message)));
 }
 
+pub fn remove_timeout(entity: Entity, timeout: &Timeout, commands: &mut Commands) {
+    if let Some(system_id) = timeout.system_id {
+        commands.unregister_system(system_id);
```
karinushka: This is a good catch! Currently it seems that the code is slowly leaking the registered system id's!
What happens is that a Timeout is added with a registered system id to run, but this system is de-registered only when the Timeout fires... if the Timeout is despawned before that, nothing is deregistering those id's...

But, the correct Bevy-like "destructor" way, would be to register a trigger, e.g. something like:
```rust
fn cleanup_timeouts(
    timeout: On<Remove, Timeout>,
    all_timeouts: ...
) {
  // de-register the system id from the above timeout.
}
```

This simplifies the rest of the code, where you don't need to worry about fetching the timeout and calling the destructor manually - just despawn the entity like it is.

src/ecs/systems.rs
```diff
@@ -422,10 +423,10 @@ pub(super) fn fresh_marker_cleanup(
     >,
     mut commands: Commands,
 ) {
-    for (entity, fresh, _) in cleanup {
+    for (entity, fresh, timeout) in cleanup {
         if !fresh {
             // Process was ready before the timer finished.
-            commands.entity(entity).try_remove::<Timeout>();
+            remove_timeout(entity, timeout, &mut commands);
```
karinushka: This commit has stray `remove_timeout`s.. they belong in a previous commit. 😺

src/ecs/systems.rs
kiwnix: The LLM agrees with you:

"I’d keep this serial. Typical managed window counts are small, so the parallel iterator setup/scheduling overhead is unlikely to pay off. More importantly, these calls are AX/macOS-facing, so keeping them serialized avoids thread-safety and ordering risks. If profiling later shows frame commits are a real bottleneck at high window counts, we can revisit with a thresholded parallel path."

src/ecs/systems.rs
kiwnix: removed the pre-drain path. The loop now pumps Cocoa first, then drains generated events afterward.

src/ecs/systems.rs
kiwnix: removed exit-time flushing of pending/coalesced mouse events. Event::Exit now writes AppExit::Success and exits directly.

src/ecs/systems.rs
kiwnix: pending mouse event is appended with received_events.extend(pending_mouse.take()), then events are delivered with messages.write_batch(received_events).

src/manager/windows.rs
kiwnix: confirmed/finalized helpers as WindowOS implementation details, not part of WindowApi.

src/manager/windows.rs
kiwnix: restored the legacy 20ms behavior, its possible that this will generate unstability or bug reports i think

src/ecs.rs
kiwnix: timeout cleanup now uses the Bevy removal path; manual stray cleanup was removed from later commits and kept in the relevant timeout lifecycle commit.

src/ecs/systems.rs
kiwnix: rewrote history so timeout cleanup changes live in fix: release lifecycle resources, not in unrelated later performance commits. Still new to llm-assisted development sorry.

kiwnix:
> Thanks a lot for looking into performance! Overall looks very good - I've left a few comments here and there, let's discuss whether some changes feel somewhat placebo like 😺
> 
> If you want you can pull the commits which are ready (without any discussion points) into a separate PR and we can push it faster.

I created a branch https://github.com/kiwnix/paneru/tree/feat/perf1-reduced with the uncommentend commits, should i create a PR from this branch?

karinushka:
I have cherrypicked some commits from this PR:
- [perf: cache layout strip membership lookups](https://github.com/karinushka/paneru/commit/d11ead23afecb6a2d1896590a53c7dc8e8ff872b) 
- [perf: use set lookup for unresolved windows](https://github.com/karinushka/paneru/commit/fe022d4b905af5ef404fb96dfdeddc3a8cc71dc1) 
- [perf: avoid duplicate AX reads in spawn logging](https://github.com/karinushka/paneru/commit/ebb2f63678ef0c967f120f74b2d49cd4da45601f) 
- [perf: improve event pump and overlay responsiveness](https://github.com/karinushka/paneru/commit/14c8e2bf17a02ad996241ae3770524cb5645ec30) 
- [perf: reduce input-path blocking and hit testing](https://github.com/karinushka/paneru/commit/3001445463f266ad81f1fa9b49dcee83700c98f3) 
- [perf: combine window frame commits](https://github.com/karinushka/paneru/commit/e568a367f0952ad59d87637973866b8308aeb2a2) 
- [perf: improve restore responsiveness](https://github.com/karinushka/paneru/commit/397bbd184a0f3b5a559b003f7775c23d20eb48a1)

Can you rebase this to `testing` again and let's see which ones are left to be merged.

karinushka:
Merged these two into `testing`:
- https://github.com/karinushka/paneru/commit/cdaef751c2001643bf3fe05280f23bcb9ca4227e
- https://github.com/karinushka/paneru/commit/aa2ef6489815754bc1eb4aa709dd04e1a5f6fc23

PR #235, opened 2026-05-18
fix: stabilize startup window layout assignment (Fix from failing tests)

kiwnix:
Fix startup layout handling so windows are not duplicated or misplaced during initialization. 

- Skip layout insertion for windows already assigned by startup reconciliation
- Avoid false native-tab detection for windows already present in a strip
- Remove an existing follower column when converting windows to tabs
- Keep ECS Position and Bounds synchronized after applying window padding
- Align test mock workspace ordering with spawned window ordering
- Add regression coverage for tab conversion without duplicate followers

Verified with cargo fmt, cargo check, cargo clippy, and cargo test.

Im not an expert in rust or macos-development, i writed this changes with the help of codex

### DIFF
--- a/src/ecs.rs
+++ b/src/ecs.rs
@@ -495,7 +495,7 @@ pub fn setup_bevy_app(sender: EventSender, receiver: Receiver<Event>) -> Result<
 }
 
 struct WindowProperties {
-    pub params: Vec<WindowParams>,
+    params: Vec<WindowParams>,
 }
 
 impl WindowProperties {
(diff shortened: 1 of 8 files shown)


karinushka:
Oof, I also have just noticed that the tests were broken - so I fixed it and now the test branch have moved. Sorry 😺, can you rebase the commit?

kiwnix:
Not necesary to rebase :), you fixed all the failing tests (with better understanding), i checked if any code in the PR improve behaviour, but looks it does not :)

PR #240, opened 2026-05-20
feat: support native macOS tab handling

kiwnix:
Group windows that macOS exposes as native tabs into a single layout slot, keep their frames synchronized, and delegate tab switching to the native system shortcut.

Preserve tab groups when moving windows across virtual spaces and add interaction coverage for focus, swap, resize, virtual move, and removal behavior.

Tested with cargo fmt --check and cargo test

Manually tested with Ghostty.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -14,7 +14,7 @@ use crate::config::Config;
 use crate::ecs::layout::{Column, LayoutStrip, StackItem};
 use crate::ecs::params::{ActiveDisplay, ActiveDisplayMut, Windows};
 use crate::ecs::{
-    ActiveDisplayMarker, ActiveWorkspaceMarker, Bounds, FocusedMarker, FullWidthMarker,
+    ActiveDisplayMarker, ActiveWorkspaceMarker, FocusedMarker, FullWidthMarker,
     NativeFullscreenMarker, SelectedVirtualMarker, SendMessageTrigger, Unmanaged, ensure_visible,
     focus_entity, reposition_entity, reshuffle_around, resize_entity,
 };
@@ -527,7 +527,9 @@ fn resize_window(
     else {
         return;
     };
-    if let Ok(mut cmds) = commands.get_entity(entity) {
+    if windows.full_width(entity).is_some()
+        && let Ok(mut cmds) = commands.get_entity(entity)
+    {
         cmds.try_remove::<FullWidthMarker>();
     }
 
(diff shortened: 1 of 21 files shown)


karinushka:
I made the detection of tabbed windows more robust in https://github.com/karinushka/paneru/commit/37db1685499a9ef692b287a7b6cc6cfd2cbdbec4
Can you rebase this to the top of `testing` and check that everything still works as intended?

kiwnix:
I rebased it to testing, the behaviour is different (the same as testing, it handles ghostty as a "normal" window, do not switch tabs when moving east/west, maybe its the better behaviour (with the old code with the patch it switched tabs when moving east/west). 

Maybe the current behaviour with your enhanced handling (testing-branch) is more standard (with the behaviour before your patch, ghostty was a bit unusable for me, but now its a "normal" window)

I can try to fix the PR and make paneru switch tabs when moving east/west, but i do not know if it is a desired/expected behaviour (the apps have their own keybindings to switch tabs). If you think the current testing is the "good" behaviour, i will close the PR :).

karinushka:
Hm.. my commit https://github.com/karinushka/paneru/commit/37db1685499a9ef692b287a7b6cc6cfd2cbdbec4 seems to be faulty. I tried opening up a new Ghostty window and it got snapped in as if it was a tab.
I need to debug a bit more with this approach of using the `CGWindowListCopyWindowInfo()` function to get info about visible windows.

Thank you very much adding all the other Tab features - I'll merge the changes in as soon as I figure out how to do the detection more robustly. Maybe you have some ideas 🤓

kiwnix:
I'm trying/going to try more :), also looking how the tabs/windows are managed in macos, looks like a quite convoluted thing hehe.

karinushka:
I seem to have fixed the detection issue in my commit.

Also I merged your commit to the top of `testing`: https://github.com/karinushka/paneru/commit/a6f248692e79e99de7242704d119d8d8699f1dc7
check it out whether everything works correctly.

kiwnix:
For me testing branch do not work correctly yet, when opening a new ghostty window, it get positioned in the same column as the old window if the size is equal (just opened, default size)

Tab handling works correctly.

I have rebased my new progress... it does not work reliably yet neither (depending on the movements with the tabs, sometimes paneru "get lost" and cannot move to a tab, also sometime when adding tabs to a ghostty that is west of another ghostty window, paneru moves the window to the east over the one on the west, but i cannot reproduce it reliably)

kiwnix:
With the changes you pushed to testing, at least with a rapid test, it works perfectly for me!!! 👏🏻 , ¿Should i close the PR?

karinushka:
> With the changes you pushed to testing, at least with a rapid test, it works perfectly for me!!! 👏🏻 , ¿Should i close the PR?

Sure. Most of the features should be in the testing branch. Check them out.
(I've reworked the unit testing harness making it simpler, so there will be large diffs because I had to rewrite a bunch of existing unit tests)

Also the feature of switching between native tabs by sending keybinds was behaving weirdly, so I have temporarily pulled it out of the `testing` and have it in my internal debug branch. I will re-merge it in once I have more thorough testing for it.

Issue #194, opened 2026-04-26
[Feature Request] Raise floating windows command


yougotwill:
OmniWM supports this feature where all floating windows are brought into focus with one command. This is useful for when I have multiple floating windows or dialogs that are behind the tile window strip.

karinushka:
Yeah, good idea! I was actually thinking of something similar, my usecase: getting mentioned during a video call while working on something... and having to scramble searching for a floating window in the back to unmute and respond 😺

PR #245, opened 2026-05-21
feat: Floating-tier focus model

### LINKED ISSUE #220: Suggestions for a better floating window mode
1. Consider adding a `window_focus_unmanaged` command to quickly surface any windows that aren't managed, hidden, or minimized. I added this feature to PaperWM.spoon a while back, and it has been really useful.
2. If a `window_focus_n/e/s/w` command is sent while an unmanaged window is focused, consider replacing the command with an action that restores focus to a managed window. Alternatively, a `window_focus_managed` command could work, if you want to avoid autonomous/DWIM-style actions.

Feel free to just close the issue if you think this is out of scope.

Thanks for your work!

### LINKED ISSUE #242: Focused-tile border remains visible behind a floating window
https://github.com/user-attachments/assets/51eb2841-581e-4e8c-b7b6-2de9de970c1f

With `[decorations.active.border]` enabled, opening a floating app (e.g. System Settings) over a focused tile leaves the focused-tile border painted behind the floating window. It looks persistent but is actually being re-painted at ~600 ms.

## Repro
1. Enable `[decorations.active.border]` and `[decorations.inactive.dim]`.
2. Focus a tile that mostly fills the screen (full-height Ghostty).
3. Open a floating app positioned to overlap that tile.
4. Watch the border behind the floating window.

Video attached.

## Cause

Focus oscillates between the tile and the floating window. `update_overlays` (`src/ecs/systems.rs:821`) correctly hides the overlay when the focused entity is `Unmanaged::Floating`, but it gets called on the tile every other tick. Instrumented traces point at `mouse.rs:157`: focus-follows-mouse transfers focus to whatever `find_window_at_point` returns in the overlap region, and `focus_without_raise` (`src/manager/windows.rs:508`) flips z-order on each transfer, so the answer keeps flipping too.

## Note

A naive guard refusing FFM out of a focused floating window stops the oscillation but kills FFM completely while any floating window is focused. Likely needs the z-order policy discussed in #220 / #194.

imrellx:
Hey, took a stab at #220 and #194. First time submitting anything this size here, so apologies up front if I've gone a direction you weren't planning on. Totally fine if this isn't the shape you'd want it in.

Would close #220, #194.

The approach borrows the floating-tier focus model from `Fjx-dylanZ/paneru` and ports the pieces I thought fit cleanly here. Three primitives, four commands.

**Primitives**
- `FocusHistory`: per-workspace `Resource` that remembers the last-focused window in each tier (managed and floating). Forgets on entity despawn and workspace destroy.
- `FloatingLayer`: per-workspace `Component` tracking whether floats sit in front of or behind the tiles. Defaults to `Front`.
- `raise_without_focus()`: trait method on `WindowApi`, implemented via `AXUIElementPerformAction(kAXRaiseAction)`.

**Commands**
- `window_focus_unmanaged`: focus the workspace's last-focused float.
- `window_focus_managed`: focus the workspace's last-focused tile.
- `window_raise_floating`: raise every visible float on the workspace, focus the last one. Idempotent on repeat presses.
- `window_togglefloatlayer`: flip the tier order, raise the new top tier, focus that tier's last-focused window.

`Focus(Direction)` now branches on the focused window. If it's a float, navigation stays in the float tier using a 45-degree cone, Euclidean nearest. Floats and tiles don't share a directional graph.

**Out of scope**

I didn't try to fix #242 here. The FFM oscillation traces back to `_SLPSSetFrontProcessWithOptions` bringing the target app frontmost on every cross-app focus, which reshuffles z-order in the overlap zones. The cleanest path I've found needs SLS-level z-order control via a `dlopen`/`dlsym` SkyLight wrapper (the pattern OmniWM uses). That feels like it deserves its own review, so I'd rather come back with it as a follow-up if you're open to it.

**Commits**

Eight feature slices plus a parser fix on top. Each one compiles and tests on its own.

**Tests**
- 5 unit tests for `FocusHistory` (record/forget/per-workspace isolation, Minimized/Hidden ignored)
- 4 unit tests for `pick_nearest_in_direction` (cone, wrong-side, First/Last return None)
- 1 integration test for the layer flip across three iterations, plus a workspace-isolation test for `window_focus_unmanaged`
- 113 tests pass. `cargo clippy --all-targets -- -D warnings` is clean apart from one `duration_suboptimal_units` at `src/config.rs:1864` that already exists on `main`.

I ran a live session of around 30 rapid toggles of `window_togglefloatlayer` across two apps and four windows. No state drift, no warnings, no regression on tile-to-tile focus. The other three commands and the float directional branch I tested less aggressively, so they might benefit from a longer soak before they ship.

No hard feelings if you'd rather close it, or if it needs reworking before it fits. Credit to `Fjx-dylanZ/paneru` for the original implementation I borrowed from.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -22,6 +22,7 @@ use crate::ecs::{
 };
 use crate::events::Event;
 use crate::manager::{Application, Display, Origin, Size, Window, WindowManager};
+use crate::platform::WorkspaceId;
 
 /// Represents a cardinal or directional choice for window manipulation.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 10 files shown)


karinushka:
Cool stuff - thanks a lot!
I was about to comment that according to Bevy best practices, the global `FocusHistory` resource should have been a component on a workspace - but then I realized that we no longer track macOS native workspaces, but multiple virtual workspaces. So this should be a refactor for the future 😺

PR #254, opened 2026-06-01
fix: consider widest window as the column width

ToroidalFox:
fixes columns overlapping due to considering only the first window as the column width

Screenshot 2026-06-01 at 14 14 46

--------------------

unrelated: auto center is broken at testing after `5d959bd`(don't know exactly when) only for _some_ columns. my initial guess is that it's about tabs.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -801,7 +801,7 @@ fn resize_window(
         .ok()
         .and_then(|idx| strip.get(idx).ok())
     {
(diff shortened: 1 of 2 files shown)


ToroidalFox:
explanations of two new commits: it allows windows within a stack to have different widths which prevents constrained windows from having wider than usual width

ungrouping from this:
Screenshot 2026-06-02 at 04 58 33

before:
Screenshot 2026-06-02 at 04 58 54

after:
Screenshot 2026-06-02 at 04 59 45

Issue #229, opened 2026-05-15
Virtual Workspace switch animation


agmitron:
Window focusing animations (horizontal) are really cool and beautiful, thanks a lot. 

It would also be a pleasure to have vertical animations, as smooth as the horizontal ones.

karinushka:
I had them when I was initially playing around with virtual workspaces, but thought that they were "too slow" and would annoy people.
So yeah, they can be behind a toggle. The only "complication" is that the inactive workspace windows are hidden in the lower bottom corner, so by default animation would show them flying out from there, which is maybe okay. Otherwise they would first need to move into position and then animate upwards.

agmitron:
Yeah, toggle is mandatory, animations are not for everyone. Also, I would consider an option to provide a "mode" in a command, e.g. `paneru send-cmd window virtual north --animated`. 

Use cases: when I "iterate" rows up and down, it would be good to see the direction where I move (because I can mess up buttons and get lost); but when I use "alt tab" to switch to the previous row, I don't need that assistance (in this case I know where I'm going and will be annoyed by transitions).

karinushka:
This is available in the latest `testing` branch: https://github.com/karinushka/paneru/commit/131e1168ff88dd2625c29f0ed55ddeab98518b6b
```toml
[options]
virtual_workspace_animations = true
```

Check it out.
The only small nit from my point of view is that the strip of windows is "flying out" from the hiding place - the lower right corner. Having a more advanced animation, where the strip flies out either from the bottom or top would require much more effort - e.g. the windows would have to be instantly moved to another position and then "fly" out in the correct direction.

agmitron:
Thanks. 

> Having a more advanced animation, where the strip flies out either from the bottom or top would require much more effort

Got it.  I thought it would be easier. 

Unfortunately, current implementation doesn't fit my use-case: for me it's not just an animation, but the direction of the row I am moving at, that helps me to understand if I go below or above.

soraxas:
> Thanks.
> 
> > Having a more advanced animation, where the strip flies out either from the bottom or top would require much more effort
> 
> Got it. I thought it would be easier.
> 
> Unfortunately, current implementation doesn't fit my use-case: for me it's not just an animation, but the direction of the row I am moving at, that helps me to understand if I go below or above.

There's a difficulty when i'm trying to implement this:
- When compare to niri on linux, because of Mac's It Just works™, there's no way for a window to move above the title bar on-top
[...]

PR #264, opened 2026-06-04
feat: optionally insert moved windows at their on-screen slot (insert_windows_mid_strip)

soraxas:
## Summary

A set of fixes and one feature around **virtual-workspace switching and moving windows between virtual workspaces**, building on the VW switch animations (#229). Together they stop windows from jumping during switches and make "move window to another VW" place the window where you actually see it.

All changes are scoped to `src/ecs/workspace.rs`, `src/ecs/layout.rs`, `src/ecs/systems.rs`, and `src/ecs.rs`, gated so existing animated behaviour is preserved.

## Problems fixed

1. **Windows jump randomly when switching/moving VW mid-animation.** If a switch or follow-move was triggered while a window or strip was still mid-animation, the intermediate position was carried into the switch — `show_active_workspace` snapshots the strip's mid-lerp position into `PreviousStripPosition` and recomposes windows from it, so windows jumped from a half-finished pose.

2. **Moving a window to another VW appended it to the end of the target strip.** It was `append`-ed to the far end, so following focus to it forced a long scroll — and the window did not land where it was on screen.

3. **With `virtual_workspace_animations` disabled, moving a window to another VW still animated.** The destination strip's windows were handed an animating `RepositionMarker` as the strip was brought on-screen and slid in over several frames, because the snap path keyed off the one-frame `ActiveWorkspaceMarker.is_added()` pulse, which can land a frame or two before the windows actually get their layout change.

4. **The moved window did not keep its on-screen x**, especially when the destination strip was scrolled or not grid-aligned — it snapped to the nearest column boundary instead of its real position.

## Functional changes

- **Settle in-flight animations before a switch/move** (`PreUpdate` keybind handlers): any pending `RepositionMarker`/`ResizeMarker` is committed to its target before the switch begins, so the switch starts from a fully-resolved layout. Done in the handlers (not a new `Update` system) to avoid perturbing the ambiguous `Update` schedule.

- **Spatial insertion when moving a window into another VW**: instead of appending, the window is inserted at the column boundary nearest its current position, and the destination strip is scrolled so the window's new slot lands **exactly** where it already was on screen — it keeps its x; the other windows shift to make room. New VW and existing-VW paths both preserve the window's x. Added `LayoutStrip::insert_tab_group_at`; `append_tab_group` now delegates to it.

- **Instant move when animations are off**: a small `ActivationSnap` countdown (set at the activation point when `virtual_workspace_animations` is disabled) makes a `PostUpdate` system settle window animation markers onto their targets for a few frames, so the destination strip appears in place instead of sliding in — regardless of the marker/layout frame timing. Only armed on an actual view switch (a "stay"/send leaves the active strip's reshuffle to animate per `animation_speed`).

- Leftover `Scrolling` on the destination strip is cleared when its scroll is set, so `apply_scrolling_constraints` doesn't override the placement with stale momentum.

## Testing

Added unit/integration tests covering: no `RepositionMarker` lingers at any update step of a move (per-frame, not after-settle); the moved window keeps its exact x into a **scrolled, non-grid-aligned** destination; new-VW follow keeps the window's x; and `insert_tab_group_at` placement. Full suite passes (142 tests), `cargo clippy` clean.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -33,6 +33,7 @@ General behavior settings for the window manager.
 | `reap_empty_workspaces` | String | `false` | If enabled, a virtual workspace without any windows will be removed. |
 | `disable_native_tabs` | Boolean | `false` | If enabled, Paneru will not auto-merge a newly-spawned window into a tab group with an existing same-app sibling that shares its frame. Use this if you find unrelated windows being grouped together. |
 | `virtual_workspace_animations` | Boolean | `false` | If enabled, Paneru will animate virtual workspace swaps. Off by default, because people use virtual workspaces due to the slow animation of the native macOS workspaces. |
+| `insert_windows_mid_strip` | Boolean | `false` | When moving a window to another virtual workspace, insert it at the column matching its current on-screen position (keeping it where you see it and shifting the rest) instead of appending it to the end of the destination strip. |
 
 ---
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -797,6 +797,14 @@ impl Config {
             .virtual_workspace_animations
             .is_some_and(|enabled| enabled)
     }
+
+    pub fn insert_windows_mid_strip(&self) -> bool {
+        // Default is disabled: appending to the end of the strip is the
+        // expected behaviour, especially when moving several windows.
+        self.options()
+            .insert_windows_mid_strip
+            .is_some_and(|enabled| enabled)
+    }
 }
 
 fn parse_hex_color(hex: &str) -> (f64, f64, f64) {
@@ -1061,6 +1069,12 @@ pub struct MainOptions {
     /// Off by default, because people use virtual workspaces due to the slow animation of the
     /// native macOS workspaces.
     pub virtual_workspace_animations: Option<bool>,
+
+    /// When moving a window to another virtual workspace, insert it at the column
+    /// matching its current on-screen position (keeping it where you see it,
+    /// shifting the rest) instead of appending it to the end of the strip.
+    /// Off by default.
+    pub insert_windows_mid_strip: Option<bool>,
 }
 
 /// Returns a default set of column widths.
--- a/src/ecs/layout.rs
+++ b/src/ecs/layout.rs
@@ -312,21 +312,30 @@ impl LayoutStrip {
     }
 
     pub fn append_tab_group(&mut self, entities: &[Entity]) {
-        let mut group = Vec::new();
-        for entity in entities {
-            if !group.contains(entity) {
-                group.push(*entity);
-            }
-        }
+        let group = dedup_entities(entities);
         if group.is_empty() {
             return;
         }
 
+        // Re-grouping existing members keeps them at their lowest current index;
+        // a foreign group lands at the end.
         let index = group
             .iter()
             .filter_map(|entity| self.index_of(*entity).ok())
             .min()
             .unwrap_or(self.len());
+
+        self.insert_tab_group_at(index, &group);
+    }
+
+    /// Inserts `entities` as a single column at `index` (clamped to the column
+    /// count), after removing any existing occurrences. A one-element group
+    /// becomes a `Single` column, more than one a `Tabs` column.
+    pub fn insert_tab_group_at(&mut self, index: usize, entities: &[Entity]) {
+        let group = dedup_entities(entities);
+        if group.is_empty() {
+            return;
+        }
 
         for entity in &group {
             self.remove(*entity);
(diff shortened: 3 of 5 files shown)


stealthybox:
Hi, just a curious, new end-user here.
I haven't tried out this PR locally,
> Spatial insertion when moving a window into another VW: instead of appending, the window is inserted at the column boundary nearest its current position

Just chiming in that this is the way I was expecting moving windows to other workspaces to work.
(mentally thinking of them as a contiguous 2d space, instead of a list of stacks)

It might have unexpected consequences though making inserts less deterministic.
Perhaps it's a different kind of move/send action, and users can pick which ones they want to use?

An edge case here is that if I'm sending 3 or 4 windows to another VW without following them, I'd expect them to insert in stack order.
When you send a window, the next one fills its existing spot at the same X coord on the left side of the window edge, so sending it to a workspace of equal length could compete for the same X position in the target VW -- in this case, both positions being equal could stack window 2 to position 2, but now on the next insert, it's impossible to stack windows 3|4 into the 3rd or 4th position, they'll always fill in in reverse order.

It behaves more predictably when the target strip is significantly longer or shorter, then the insert order flips to reversing or stacking since it's flowing spatially/positionally until the target strip passes the previous one in length.

I don't think this is necessarily a bad behavior, but it's worth talking through since the current one always does the same thing, no matter how long strips are relative to each other. A third behavior would be always inserting at the beginning. Maybe this could be an option or another movement/send type, since there's already similar window swapping actions for the front and back of the strip.

I'm running a local build with some personal changes that allow using the send-cmd cli action to move windows by ID between workspaces without changing the active focus:
```
      paneru send-cmd window focusid <id>
      paneru send-cmd window virtualmovenumid <id> <n>
      paneru send-cmd window virtualsendnumid <id> <n>
```
They let scripts drive focus and virtual-workspace placement.
I'm using it to change focus with a window fuzzy-finder script with `fzf`, and another script for moving all my chat apps and music to workspace 5.
I was thinking of adding another option that supports specific window insertion index.
```
      paneru send-cmd window virtualmovenumidindex <id> <n> <i>
      paneru send-cmd window virtualsendnumidindex <id> <n> <i>
```
I'll see if I can find some time this week to try your change to the insertion behavior out.

karinushka:
Check out the two small commits I added to `testing`:
- https://github.com/karinushka/paneru/commit/ea2f3762d4cebe129a4b34d678bb22db900e149a should fix the issue with position being saved mid-animation. Done somewhat simpler, without adding additional systems which keep track with global flags and run additional systems.
- https://github.com/karinushka/paneru/commit/dfba539422dfabf7cfd0c809e075c6e144bd1bd2 I realized that detection of workspace switching can be done very simply - by looking at the vertical `y` changes. If the windows are being moved vertically they are being hidden/showed in a workspace change. The commit is also relatively simple, removing a bunch of tracking code.

Check these two out and see whether they fix the issues you found!

Regarding the feature of positioning a moved window mid-strip, I somewhat agree with @stealthybox - it can be a bit confusing that a moved window ends up in the middle of your existing strip, breaking the order.
Either way the move operation is the same: you move a window to another strip and then shuffle it into the correct position. Having it appear at the end of the layout strip "feels more natural", especially when moving several windows.

Otherwise we can toggle this behavior behind a config flag. In such case a clean and neat implementation would be to track the `VirtualMoveMarker`. When this marker is removed this signals that a window has been moved to another virtual workspace - so we can run a dedicated system which takes the window and positions the strip so the window is at its current position.
Having it in a separate systems also allows for a neat built-in `run_if(config.insert_windows_mid_strip())` condition.

soraxas:
> Regarding the feature of positioning a moved window mid-strip, I somewhat agree with @stealthybox - it can be a bit confusing that a moved window ends up in the middle of your existing strip, breaking the order.

It's been a while since i have my linux box, but IIRC i think niri does insert current active window to the new strip mid-location, instead of appending? (i think that's why i expect it to behave like this).

- Do you think its worth making it more like a generic command that can be insert in any location `n`, where `0` is always the left-most and `-1` is the right most, and anything in-between would be the location accordingly.
- Or would that be an overkill?
 
> ...it can be a bit confusing 

i mean, appearing on the RIGHT-most of strip (instead of LEFT-most) is already an assumption that things should goes left-to-right.

karinushka:
Ok, I can see both sides of the argument. If you want, re-fresh this PR or make a new one which adds the "insert window at the current slot when switching virtual workspaces".

Nit: that `handle_virtual_window_moves()` function is huuuuge. I wonder whether it can be logically split up into a couple of independent systems, all reacting to the marker.

soraxas:
Refreshed the PR. Summary of what changed in response to the review:

- **Dropped the animation/jump machinery.** Your two commits on `testing` (destination-capture into `PreviousStripPosition`, and the vertical-`y` workspace-switch detection) handle it — I rebased onto them and confirmed the jump and the instant-switch-when-animations-off both hold *without* the extra `ActivationSnap` resource / pre-switch snap systems. Agreed yours is the cleaner approach; removed mine.

- **Mid-strip insertion is now opt-in** behind `insert_windows_mid_strip` (default off), so append-to-end stays the default — matches what you and @stealthybox preferred.

- **Did it as a dedicated system**, per your suggestion: `place_moved_window_in_slot` with `run_if(config.insert_windows_mid_strip())`, reacting to the moved window after `handle_virtual_window_moves`. So `handle_virtual_window_moves` is untouched

yougotwill:
@soraxas This setting doesn't work when moving windows between workspaces on different displays using the `window_nextdisplaysend`?

Issue #274, opened 2026-06-11
[Feature] `window_balance` command


yougotwill:
I would like a command similar to `window_equalized` but to make all columns on a strip a set width. This would be useful for resetting layouts.

I'm not sure if you would prefer to make a separate variable for what the set width should be or make it similar to the resize commands that use `preset_column_widths`

An example use case would be when I unplug my monitor I want to quickly make all the windows on a strip full width because for me it's easier to navigation on a smaller screen.

You could call this `resize_all` or `grow_all` etc.

karinushka:
Yup, sounds like a good idea - I think the simplest yet still ergonomic option would be to have a command which takes the with of the current window and makes all the other windows in the same strip same width.
What do you think?

yougotwill:
> Yup, sounds like a good idea - I think the simplest yet still ergonomic option would be to have a command which takes the with of the current window and makes all the other windows in the same strip same width. What do you think?

Sounds good to me 👍

PR #280, opened 2026-06-17
feat: Add window balance command to equalize column widths

yougotwill:
Resolves https://github.com/karinushka/paneru/issues/274

- Add a window balance command that takes the width of the currently focused window and applies that same width to all columns in the active strip, then reshuffles the layout.

Edge cases handled:
- No focused window → early return.
- Focused window not in a strip → early return.
- Fullscreen columns are skipped.
- Full-width markers are cleared so the resize takes effect.
- WidthRatio is automatically updated by commit_window_size on the next frame when Bounds changes.

Test design:
- 3 windows with default config (each 400 px wide on a 1600 px display).
- Resize(Grow) once on window 0 → width becomes 512 px (25% → 50% of 1024 px viewport).
- Balance command.
- Assert window 0, 1, and 2 are all 512 px wide.

**Note: This PR was built with AI but reviewed and tested by me.**

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -139,6 +139,7 @@ Format: `"[modifiers-]key"`. Available modifiers are:
 | `window_stack` | Stack the current window into the column on the left. |
 | `window_unstack` | Pull a window out of a stack into its own column. |
 | `window_equalize` | Make all windows in a stack equal height. |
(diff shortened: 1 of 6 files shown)


yougotwill:
I'm new to rust so any feedback and guidance on best practices would be appreciated :)

karinushka:
Yup, looks good.
Thank you for your contribution!

PR #284, opened 2026-06-24
Forced window management

### LINKED ISSUE #230: [Bug] paneru does not tile BetterTouchTool
Using the `testing` branch https://github.com/karinushka/paneru/commit/4cbe095fbf7b0e0c2835f93ffb0313dbf6389906.

Whenever I open the main window for BTT it is always floating. Using the floating toggle doesn't change it's state to tiling. Trying to assign a window rule for the main window doesn't seem to work either.

I do have a separate window rule for the screenshot window but confirm even if I remove it the main window still doesn't tile.

```toml
[windows.floating_bettertouchtool_screenshots]
bundle_id = "com.hegenberg.BetterTouchTool"
title = "Screenshot*"
floating = true
```

yougotwill:
Fixes #230 

Introduces a `manage` boolean option to force Paneru to manage background applications (`LSUIElement`) or non-standard windows that are normally ignored. This allows users to properly manage apps like BetterTouchTool, Chronoid or ProtonVPN by explicitly opting in.

Example:

```toml
[windows.bettertouchtool]
bundle_id = "com.hegenberg.BetterTouchTool"
title = ".*"
manage = true
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -404,28 +404,19 @@ impl Config {
             })
     }
 
-    /// Finds window properties for a given `title`, `bundle_id`, `role`, and `subrole`.
+    /// Finds window properties for a given `title` and `bundle_id`.
     /// It iterates through configured window parameters and returns all matching rules.
-    /// A rule matches when its bundle ID (if any) and title regex match, and when any
-    /// optional `role`/`subrole` regexes also match.
+    /// A rule matches when its bundle ID (if any) and title regex match.
     ///
     /// # Arguments
     ///
     /// * `title` - The title of the window to match.
     /// * `bundle_id` - The bundle identifier of the application owning the window.
-    /// * `role` - The accessibility role of the window (optional).
-    /// * `subrole` - The accessibility subrole of the window (optional).
     ///
     /// # Returns
     ///
     /// A `Vec<WindowParams>` containing all matching window rules.
-    pub fn find_window_properties(
-        &self,
-        title: &str,
-        bundle_id: &str,
-        role: Option<&str>,
-        subrole: Option<&str>,
-    ) -> Vec<WindowParams> {
+    pub fn find_window_properties(&self, title: &str, bundle_id: &str) -> Vec<WindowParams> {
         self.inner()
             .windows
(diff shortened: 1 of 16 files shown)


src/config.rs
```diff
@@ -433,6 +434,20 @@ impl Config {
             .unwrap_or_default()
     }
 
+    /// Returns `true` if any window rule for the given bundle ID requests that the
+    /// process be forcibly managed even when macOS reports it as unobservable.
+    pub fn should_force_manage_process(&self, bundle_id: &str) -> bool {
```
karinushka, line 439: nit: Probably better to pass a `&Process` as parameter here, instead of `bundle_id`. This way the function can extract the `bundle_id` itself, instead of the callers having to do it every time. Should simplify the call points somewhat.

karinushka (approved):
Cool stuff!
Thanks for looking into it - you have a wide range of "non-compliant" apps, so if it works for you now, it should work everywhere 😺

Looks good overall, just a small nit simplifying calling points for a common function. It seems like if the underlying architecture is simple and clean, LLM agents have no trouble producing reasonable output.

karinushka:
I cleaned it up slightly and committed this with you as the author - https://github.com/karinushka/paneru/commit/d95b4d811d78807d7b768599b00223ae116ba16d
Check it out and see whether everything works as expected.

yougotwill:
> I cleaned it up slightly and committed this with you as the author - [d95b4d8](https://github.com/karinushka/paneru/commit/d95b4d811d78807d7b768599b00223ae116ba16d) Check it out and see whether everything works as expected.

Thanks for this @karinushka! Sorry I have been absent the last few weeks. I am running it and will report back any issues. So far seems to be working in most cases! Particularly for BetterTouchTool and Chronoid as previously mentioned in the issue #230.

One interesting I learned when using regex's for floating windows. I was trying to capture windows that do not have a title (an example would be the pomodoro break screen with Chronoid) is that using `\A\z` is more effective than using `^$` for whatever reason. I do sometimes find the apps are still not detected in some cases.

Issue #291, opened 2026-07-07
`continuous = false` config option unexpected behaviour


ivanbalashov239:
## Current behavior

The swipe gesture and scroll-wheel sliding are always **smooth/continuous** — you can stop mid-way between windows, and the strip scrolls freely. The `[swipe] continuous` option (currently implemented in `src/ecs/scroll.rs:298-324`) only controls **edge clamping** (whether you can overscroll past the first/last window), not the snap behavior:

- `continuous = true`: allows scrolling past the edge windows (overscroll)
- `continuous = false`: clamps the strip so edge windows stay on-screen

In both cases the scrolling itself remains smooth and free-form.

## Expected behavior

Like Niri, when scrolling horizontally (via trackpad swipe or scroll wheel), the viewport should snap **window-by-window** — each scroll increment brings one full window into view at a predictable position. This allows muscle-memory scrolling where you know exactly how many swipes/scrolls it takes to reach a particular window.

Not sure if its actually expected behaviour, or my personal desire, but this are my two cents

karinushka:
Yeah, there is a slight confusion in the naming ☺️
I also recently noticed that Niri had a similarly named option.

I think the actual `continuous = false` should be implemented as a focus shift, to bring the window fully into view.

Issue #80, opened 2026-02-08
When installing via `nix` + `home-manager`, `paneru` closes on launch if no config file is found.


josh-sukheja:
I'm not sure if this is a nix-specific problem, or if it's just how paneru works.

I followed the project readme and have installed paneru like this:
```nix
# paneru.nix, imported by flake.nix

{ inputs, ... }:

{
  imports = [ inputs.paneru.homeModules.paneru ];

  services.paneru = {
    enable = true;
  };
}
```
Note that I have not defined a `services.paneru.settings = { ... }` block. I also do not have a paneru config file on my system (fresh install).

The project readme documents the default config options if none are specified. I assumed this means if no config file is present at all, then I would get all defaults. But it seems I must bring my own config file to start paneru.

These are the errors I see in `/tmp/paneru.err.log`:
```
# bat /tmp/paneru.err.log:

   1 │ 
   2 │ thread 'main' (7482) panicked at src/config.rs:60:13:
   3 │ paneru::config::CONFIGURATION_FILE::{{closure}}::{{closure}}: Configuration file not found. Tried: $PANERU
     │ _CONFIG, $HOME/.paneru, $HOME/.paneru.toml, $XDG_CONFIG_HOME/paneru/paneru.toml
   4 │ note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I can fix this by changing the nix config to this, which generates an empty config file:
```nix
services.paneru = {
  enable = true;
  settings = { };
};
```

agmitron:
https://github.com/karinushka/paneru/issues/64#issuecomment-3974233762

This might temporarily help you, but actually it should be working out of the box.

karinushka:
Its in `main` now. If Paneru starts without a config, it will create a default one.

PR #300, opened 2026-07-15
[WIP / NOT READY] Oversized windows and native width controls

MrFlashAccount:
> [!CAUTION]
> **WIP — NOT READY FOR REVIEW OR MERGE**
>
> This branch has been rewritten into a focused upstream slice and retargeted to `testing`. The behavior is working, but the implementation still needs an upstream design decision and one cleanup pass.

## Intent

Allow Paneru-managed windows to be wider than the visible display, expose those widths through the native menu bar, and keep focused floating windows manageable.

## Current scope

- accepts width ratios above `1.0`
- exposes configured width presets in the menu bar
- retries oversized resizes when macOS initially constrains the requested frame
- allows the focused floating window to enter managed mode
- includes focused regression coverage for a two-display-width window

## Known blockers before ready

- decide whether the retry belongs only in the low-level resize path; the current upper-level retry may be redundant
- confirm the desired maximum width and configuration contract for upstream
- manually verify behavior across multiple displays and applications with strict resize constraints

## Verification

- `cargo check --all-targets`
- `cargo test tests::tiling::test_window_can_resize_to_two_display_widths_and_scroll -- --exact`
- `cargo test menubar::tests -- --test-threads=1`

## Explicitly excluded

- gesture/momentum changes
- sticky or paged scrolling
- event-driven runtime and persistence
- Accessibility onboarding
- Sparkle/release infrastructure

### DIFF
--- a/src/manager/windows.rs
+++ b/src/manager/windows.rs
@@ -38,6 +38,25 @@ use crate::util::{AXUIAttributes, AXUIWrapper, MacResult};
 static ENHANCED_UI_REFCOUNT: LazyLock<Mutex<HashMap<Pid, usize>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
+/// macOS may partially apply an AX width increase when the requested right edge
+/// would be far outside the display. Moving the partial result left by the
+/// missing width before retrying gives `WindowServer` enough offscreen room.
+///
+/// Only retry when the first attempt actually grew the window. Fixed-size apps
+/// otherwise look identical to this failure mode and must not be moved offscreen.
+fn resize_staging_origin(
+    previous_frame: IRect,
+    actual_frame: IRect,
+    target_width: i32,
+) -> Option<Origin> {
+    let actual_width = actual_frame.width();
+    (actual_width > previous_frame.width() && actual_width < target_width).then(|| {
+        actual_frame
+            .min
+            .with_x(actual_frame.min.x - (target_width - actual_width))
+    })
+}
+
 #[derive(Debug)]
 pub enum WindowPadding {
     Vertical(i32),
@@ -313,6 +332,56 @@ impl WindowOS {
         }
     }
 
+    fn set_ax_position(&mut self, origin: Origin) {
+        let mut point = CGPoint::new(
+            f64::from(origin.x + self.horizontal_padding),
+            f64::from(origin.y + self.vertical_padding),
+        );
+        let position_ref = unsafe {
+            AXValueCreate(
+                kAXValueTypeCGPoint,
+                NonNull::from(&mut point).as_ptr().cast(),
+            )
+        };
+        if let Ok(position) = AXUIWrapper::retain(position_ref) {
+            unsafe {
+                AXUIElementSetAttributeValue(
+                    self.ax_element.as_ptr(),
+                    CFString::from_static_str(kAXPositionAttribute).as_ref(),
+                    position.as_ref(),
+                )
+            };
+            let size = self.frame.size();
+            self.frame.min = origin;
+            self.frame.max = origin + size;
+        }
+    }
+
+    fn set_ax_size(&mut self, size: Size) {
+        let width_padding = 2 * self.horizontal_padding;
+        let height_padding = 2 * self.vertical_padding;
+        let mut cgsize = CGSize::new(
+            f64::from(size.x - width_padding),
+            f64::from(size.y - height_padding),
+        );
+        let size_ref = unsafe {
+            AXValueCreate(
+                kAXValueTypeCGSize,
+                NonNull::from(&mut cgsize).as_ptr().cast(),
+            )
+        };
+        if let Ok(size_value) = AXUIWrapper::retain(size_ref) {
+            unsafe {
+                AXUIElementSetAttributeValue(
+                    self.ax_element.as_ptr(),
+                    CFString::from_static_str(kAXSizeAttribute).as_ref(),
+                    size_value.as_ref(),
+                )
+            };
+            self.frame.max = self.frame.min + size;
+        }
+    }
+
     /// Makes the window the key window for its application by sending synthesized events.
     ///
     /// # Arguments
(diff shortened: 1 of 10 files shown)


karinushka:
Oof.. I missed those unit tests. Fix it now in the main branch. Thanks for pointing it out!

Also thank you for contributing the PR - I never thought about having windows wider than the screen, but it actually makes good sense. Especially using the swiping gestures, one can have an extra wide window and it still can be quite useable on a smaller laptop screen!

src/manager/windows.rs
```diff
@@ -38,6 +38,25 @@ use crate::util::{AXUIAttributes, AXUIWrapper, MacResult};
 static ENHANCED_UI_REFCOUNT: LazyLock<Mutex<HashMap<Pid, usize>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
+/// macOS may partially apply an AX width increase when the requested right edge
+/// would be far outside the display. Moving the partial result left by the
+/// missing width before retrying gives `WindowServer` enough offscreen room.
+///
+/// Only retry when the first attempt actually grew the window. Fixed-size apps
+/// otherwise look identical to this failure mode and must not be moved offscreen.
+fn resize_staging_origin(
```
karinushka, line 47: Yeah, good idea to handle this macOS peculiarity at the lower level - essentially hiding this from above layers.
This should make some code higher up redundant - currently when resizing a partially offscreen window it is first moved into view.

src/config/swipe.rs
```diff
@@ -23,6 +23,10 @@ pub struct SwipeOptions {
     #[allow(dead_code)]
     pub continuous: Option<bool>,
 
+    /// Snap to the nearest column after native momentum settles.
```
karinushka: Ooh, this is very similar to the feature request https://github.com/karinushka/paneru/issues/291
Append `(#291)` to the commit name for greater justice 😺

karinushka (commented):
All the `fix:` commits are great - thanks for helping out!
If possible rebase them onto `testing` branch as the base. That way we can bake them for a bit before promoting into the `main` branch.

The Sparkle auto-update changes are complicated - maybe pull them out into a separate PR so we can discuss.
The current "release procedure" is: commits arrive into the `testing` branch, and then the trailing `main` branch slowly catches up with them - this allows the changes to be tested before unleashed on everybody. Every once in a while I do a release cut on the main branch and sync the version to Crates and Brew.

MrFlashAccount:
@karinushka thank you for reviewing this! Yeah, as it's a WIP pr, take the changes with a grain of salt, I'll open a PR once I'm satisfied with the result. For sure, I'll exclude the release workflow from the pr, since the changes are unrelated. 

FWIW, if you wanna test the prototype, download it here: https://github.com/MrFlashAccount/paneru

Your feedback is always welcome!

MrFlashAccount:
Quick status update from the fork: this started as an oversized-width experiment, but the fork has since evolved into a fairly opinionated variant of Paneru. The current fork `main` is now 33 commits ahead of upstream, so this WIP PR no longer represents the full set of changes. I also do not think merging the whole fork as one PR would be appropriate.

The most important product-level difference is **per-window, opt-in management**:

- Normal application windows now start in passthrough mode and behave like ordinary macOS windows.
- The user must explicitly choose **Toggle Managed** for each concrete window, either from the menu bar, with a shortcut, or through a matching `[windows]` rule with `manage = true`.
- Managing one window does not automatically manage every other window belonging to the same application.
- Toggling management off restores the window's pre-management frame when one was captured.
- Layout, scrolling, persistence, decorations, and cleanup are kept away from passthrough windows.

This is intentionally a substantial change from the original automatic window-manager behavior. It makes the fork feel more like an opt-in horizontal workspace that can coexist with ordinary macOS windows, rather than a manager that immediately takes ownership of all eligible windows.

Other notable changes currently present in the fork include:

1. **Oversized windows and explicit width presets**
   - Windows can be 150%, 200%, or another configured percentage of the display width.
   - Exact widths are available from native menu-bar items and keyboard shortcuts.
   - Oversized windows remain pannable instead of being clamped back to the display width.

2. **Reworked strip scrolling**
   - Modifier-scroll support with native trackpad momentum.
   - Configurable sensitivity and direction.
   - One-hop paging between adjacent window-edge stops.
   - Sticky edge snapping with a configurable `snap_padding`.
   - Correct handling of multiple windows and multiple oversized windows in the same strip.
   - Unconfigured macOS gestures are left available to the system.

3. **Menu-bar application UX**
   - Native shortcut glyphs are shown next to menu items.
   - The focused window can be managed, resized, or centered from the menu bar.
   - The application remains a menu-bar app without a Dock window.

4. **Session persistence**
   - Managed layouts, virtual workspace rows, display assignments, oversized width ratios, and horizontal strip positions can be restored after restart.
   - Persistence is debounced and disabled completely when restore is turned off.

5. **Runtime and animation work**
   - The runtime is event-driven instead of continuously polling while idle.
   - Animation pacing follows the display refresh cadence.
   - Accessibility observation is scoped and reconciled instead of permanently observing every irrelevant background application.

6. **Accessibility and first-launch behavior**
   - Missing Accessibility permission no longer produces an endless prompt loop.
   - A menu-bar onboarding state explains how to manually add the installed app in System Settings.
   - A default config is created automatically and config changes are reloaded while Paneru is running.

7. **Fork-specific distribution**
   - The fork has a packaged `.app`, DMG releases, Sparkle updates, and its own release workflow.
   - I assume this part should remain separate from upstream unless you explicitly want to discuss it.

There are also smaller fixes around native scroll momentum, constrained resize retries, virtual workspace behavior, multi-display handling, native tabs, and startup restore.

Which parts of this direction, if any, would you like to bring upstream?

I can split selected work into focused PRs rebased onto `testing`, for example:

- isolated gesture, resize, and lifecycle fixes;
- oversized width support and width menu items;
- paging and sticky scrolling;
- per-window opt-in management;
- persistence and event-driven runtime work;
- Accessibility and menu-bar UX.

I would keep the Sparkle/release infrastructure out of those PRs unless requested. The per-window opt-in behavior is probably the most important decision to make explicitly, because it changes Paneru's default ownership model rather than merely adding a feature.

karinushka:
Thanks for taking a stab at this! Your "generate default config" was an old feature request (#80) so I merged it into `testing` with you as the author (https://github.com/karinushka/paneru/commit/a47bb927c9e3d354cd4bd26d7277afb7a99f24db).

> * isolated gesture, resize, and lifecycle fixes;
> * oversized width support and width menu items;
> * paging and sticky scrolling;
> * persistence and event-driven runtime work;
> * Accessibility and menu-bar UX.

These are quite self contained - so you can rebase them on top of `testing`. I copied them from this PR onto my local testing branch and they are working nicely.
The event-driven runtime is a very interesting one - whether you managed to make the current polling more efficient, while keeping the responsiveness. There was a fork at some point where somebody tried various approaches to replace the current event pumping approach, but it resulted in choppy animations.

> * per-window opt-in management;

This one is good, but changes the expected default behavior - a good way would be to gate it behind a config flag. E.g. when Paneru starts it will by default try to manage all the windows, unless an "opt in management" flag is enabled. Then it simply floats all the windows.

Regarding the release infrastructure - Paneru is installed from source, because this avoid the "unsigned binary" warning from macOS. Do you have the iOS developer subscription to sign the releases in the github workflows?

karinushka:
Manually merged as:
* 25f56afb0292f8fb691188866ee8de598739a4b9
* 7cb04ac3568913d3c5be6f228393bba2802769e7
* 6fe6b4383dda805c9776a8befd2fafacbb1e130e

PR #303, opened 2026-07-16
Add support for Fn/Globe key as a modifier

### LINKED ISSUE #193: Using Globe key as a modifier
Hi, I am trying to use the Globe key (Fn key) as part of the modifiers, but it does not seem to be supported according to the source code. Would you consider supporting Globe key? Thank you.

Matt343:
fixes #193 

# Summary

Add support for the Fn/Globe key on MacOS keyboards as a modifier key. There's no left/right split for this key so it doesn't have the `lfn` `rfn` variants. I expanded the Modifiers type to  a `u16` since the bitmask was full already. There's an issue where macos reports `fn - f3` style keys as the hardware media keys without the fn modifier mask set if you use fn to access the function keys, but you can still bind that combination as just plain `f3` and it should read anyway, the OS handles those remappings before it emits the event.

# Testing

Added fn to the unit tests where it made sense, and then built and tested on my laptop to make sure the key detection works as expected.

I used an agent to help write the code, I don't know if you care but some projects do, all of the changes were reviewed and tested by a human on an M3 Pro.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -119,6 +119,7 @@ Format: `"[modifiers-]key"`. Available modifiers are:
 - `ctrl`, `lctrl`, `rctrl`
 - `cmd`, `lcmd`, `rcmd`
 - `shift`, `lshift`, `rshift`
+- `fn`
 
 | Action | Description |
 | :--- | :--- |
--- a/src/config.rs
+++ b/src/config.rs
@@ -1364,6 +1364,7 @@ fn parse_modifiers(input: &str) -> Result<Modifiers> {
             "ctrl" => Modifiers::CTRL,
             "lctrl" => Modifiers::LCTRL,
             "rctrl" => Modifiers::RCTRL,
+            "fn" => Modifiers::FN,
             _ => {
                 return Err(Error::InvalidConfig(format!(
                     "{}: Invalid modifier: {modifier}",
@@ -1678,6 +1679,7 @@ fn test_virtual_keymap() -> Vec<(String, u8)> {
 
 #[test]
 #[allow(clippy::float_cmp)]
+#[allow(clippy::too_many_lines)]
 fn test_config_parsing() {
     let input = r#"
 [options]
@@ -1688,6 +1690,7 @@ quit = "ctrl+alt-q"
 window_manage = "ctrl+alt-t"
 window_stack = ["ctrl-s", "alt-s"]
 window_shrink = "alt-d"
+window_snap = "fn-x"
 
 [windows]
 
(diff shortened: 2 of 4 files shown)


karinushka:
Cool stuff! Thanks for helping out - that was a long standing feature request ☺️

Agentic code is fine, as long as the person guiding knows what they are doing and most importantly it's reviewed. As the famous Mitchell Hashimoto said: "I read the code".

Do you mind rebasing it on top of the`testing` branch - that way we reduce some force pushes.

Matt343:
Rebased and squashed. Heard you on the reviews and will not send anything over in a PR that I haven't read and tested myself, I know I hate when people do that at work

josh-sukheja:
Drive by: I saw this and just updated to try it out. My very initial report: It works, but some keys are "sticky". Some of my keybinds:

```toml
window_focus_west = "fn - leftarrow"
window_focus_east = "fn - rightarrow"
window_grow = "fn - r"
```

Now after multiple `paneru restart`s, the behaviour on my inbuilt keyboard is:
- Simply pushing left/right arrows will swap window focus. My arrow keys on their own do not actually move the cursor in text fields. Holding `Fn` + arrow keys will not move the windows, it will move the cursor. So here it feels like `Fn` is stuck "on"?
- The opposite behaviour for the `window_grow` command: Pressing `r` is normal, and I need to press `Fn + r` to grow the window. This is what I expect.

I have this Macos option off, before and after, but turning it on did not seem to change the behaviour:

image

I haven't tested other commands yet because it's actually quite confusing right now. However, I did roll back to my old `alt + ...` keybinds, and I found that some of them just didn't even work anymore. I actually just rolled back to my last install, which is on 01b95a918e6942154138bbc9d8ec9dc298fb30e6, and they work again. Is something broken on `main` maybe?

Sorry that this is not an Issue, I just thought I'd put it here on the relevant PR for now.

manpreetmaso:
Is there any possibility this may have broken the CMD key entirely?

PS. Maybe on keyboards with remapped cmd / opt keys.

karinushka:
I debugged it a bit and it's actually quite interesting! @josh-sukheja 
When you press `fn - leftarrow` the macOS actually converts it to `home` and sends it as `fn - home`. At least on my macbook.

Try to do:
```toml
window_focus_west = "fn - home"
window_focus_east = "fn - end"
```
and see whether it works as expected.

Also it seems to have weird interactions with other modifier keys - I'll have a commit which checks for `Fn` key separately.

karinushka:
There should be a fix on the `testing` branch: https://github.com/karinushka/paneru/commit/97305d854d9891d5c066a79ade7ce7850296baf1
Can yo check whether it fixes it? Then I'll expedite it into the `main`.

karinushka:
@josh-sukheja The fix should be in the `main` branch now.

PR #305, opened 2026-07-16
[WIP / NOT READY] Remove managed-window gaps during native fullscreen

MrFlashAccount:
> [!CAUTION]
> **WIP — NOT READY FOR REVIEW OR MERGE**
>
> This is an independent lifecycle fix with focused regression coverage.

## Intent

Stop a managed window from leaving a reserved gap in its original layout strip after macOS moves that window into native fullscreen.

## Root cause

The layout could retain stale membership when focus markers no longer represented the window's actual macOS Space/fullscreen state. Unmanaging the window happened to clear the gap, but should not be required.

## Current scope

- reconciles managed layout membership against actual Space/fullscreen state
- removes a fullscreen window from the old strip
- restores normal participation when the window returns
- adds regression coverage for the transition

## Known blockers before ready

- manual verification with native fullscreen entry/exit on multiple Spaces
- confirm behavior for applications that report Accessibility fullscreen state late
- confirm multi-display behavior with “Displays have separate Spaces” enabled and disabled

## Verification

- `cargo check --all-targets`
- `cargo clippy --bin paneru -- -D warnings`
- `cargo test tests::interaction::native_fullscreen_window_is_removed_from_strip_and_restored_after_exit -- --exact`

## Explicitly excluded

- oversized widths and scrolling behavior
- runtime/persistence changes
- Accessibility onboarding and release infrastructure

### DIFF
--- a/src/ecs/workspace.rs
+++ b/src/ecs/workspace.rs
@@ -85,6 +85,36 @@ pub(crate) struct PreviousStripPosition {
     pub focus: Option<Entity>,
 }
 
+fn fullscreen_window_in_strip(
+    workspace_id: WorkspaceId,
+    strip: &LayoutStrip,
+    windows: &Windows,
+    window_manager: &WindowManager,
+) -> Option<Entity> {
+    window_manager
+        .windows_in_workspace(workspace_id)
+        .ok()
+        .and_then(|window_ids| {
+            window_ids.into_iter().find_map(|window_id| {
+                windows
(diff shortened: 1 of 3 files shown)


karinushka:
Merged this manually because there was a merge conflict: https://github.com/karinushka/paneru/commit/7dd98e02ccd82ec6fa9bda5ac96c0daad2ab3622

PR #307, opened 2026-07-16
One-hop paging with stable edge snapping

MrFlashAccount:
> [!NOTE]
> The maintainer-reported workspace and wide-display edge cases now have deterministic regression coverage and are fixed in this branch.

## Summary

Adds predictable one-hop horizontal paging for regular and oversized windows.

- paging derives its neighborhood from the actual strip position and the nearest global stops
- each gesture is constrained to one adjacent stop, including when several windows are visible on a wide display
- changing native macOS workspaces clears stale scrolling and reposition state from the previous strip
- an unsnapped strip cannot skip the first stop in either direction
- edge snapping activates within 10% of the usable viewport width
- physical-scroll interpolation is committed at native momentum boundaries, avoiding an extra non-native decay tail

## Configuration

```toml
[swipe]
paging = true
sticky = false
```

`paging = true` is the default. With `paging = false`, `sticky = true` enables edge-only settling without one-hop constraints.

## Verification

- `cargo test --locked --all-targets -- --test-threads=1` — 189 passed
- `cargo clippy --locked -- -D warnings`
- regression coverage for native workspace transitions, semi-offscreen windows on wide viewports, unsnapped origins, native momentum boundaries, oversized columns, and free-scroll layout tests

## Pre-merge manual validation

- trackpad feel on a laptop and an external display

## Out of scope

- event-driven runtime and persistence
- Accessibility onboarding
- release infrastructure

### DIFF
--- a/src/ecs/scroll.rs
+++ b/src/ecs/scroll.rs
@@ -63,8 +63,10 @@ fn swipe_gesture(
 ) {
     let swipe_sensitivity = config.swipe_sensitivity();
     let mut total_delta = 0.0;
+    let mut gesture_delta = 0.0;
     let mut touchpad_down = false;
     let mut has_scroll_event = false;
+    let mut has_gesture_event = false;
 
     // Normalization: Touchpad deltas are typically small fractions.
     // Scroll wheel deltas can be larger. We scale it down slightly
@@ -88,10 +90,12 @@ fn swipe_gesture(
             Event::Swipe { delta, fingers }
                 if config
                     .swipe_gesture_fingers()
-                    .is_none_or(|fingers_configured| fingers_configured == *fingers) =>
+                    .is_some_and(|fingers_configured| fingers_configured == *fingers) =>
             {
                 total_delta += delta;
+                gesture_delta += delta;
                 has_scroll_event = true;
+                has_gesture_event = true;
             }
             _ => (),
         }
@@ -117,15 +121,21 @@ fn swipe_gesture(
         };
 
         let dt = time.delta_secs_f64();
-        let new_velocity = if dt > 0.0 {
-            total_delta * swipe_sensitivity / dt
+        let new_velocity = if has_gesture_event && dt > 0.0 {
+            gesture_delta * swipe_sensitivity / dt
         } else {
             0.0
         };
 
         if let Some(scrolling) = scrolling.as_mut() {
-            // Smoothen velocity changes using EMA.
-            scrolling.velocity = 0.3 * new_velocity + 0.7 * scrolling.velocity;
+            // Native modifier-scroll events already include macOS momentum.
+            // Add synthetic inertia only for raw multi-finger gestures.
+            scrolling.velocity = if has_gesture_event {
+                // Smoothen gesture velocity changes using EMA.
+                0.3 * new_velocity + 0.7 * scrolling.velocity
+            } else {
+                0.0
+            };
             scrolling.is_user_swiping = true;
             scrolling.last_event = Instant::now();
             scrolling.position +=
@@ -135,7 +145,7 @@ fn swipe_gesture(
                 velocity: new_velocity,
                 position: f64::from(position.0.x)
                     + total_delta * viewport_width * direction_modifier * swipe_sensitivity,
-                is_user_swiping: touchpad_down,
+                is_user_swiping: true,
                 last_event: Instant::now(),
             });
         }
@@ -386,7 +396,7 @@ fn vertical_swipe_gesture(
             Event::VerticalSwipe { delta, fingers }
                 if config
                     .swipe_gesture_fingers()
-                    .is_none_or(|fingers_configured| fingers_configured == *fingers) =>
+                    .is_some_and(|fingers_configured| fingers_configured == *fingers) =>
             {
                 state.last_event = Some(Instant::now());
 
--- a/src/ecs/scroll.rs
+++ b/src/ecs/scroll.rs
@@ -156,7 +156,7 @@ fn swipe_gesture(
     // AppKit can report physical Ended and momentum Began together. Apply the
     // physical end first so the momentum phase remains the final state.
     mark_physical_touch_end(touchpad_physical_up, scrolling.as_deref_mut());
-    resume_touchpad_gesture(resumes_gesture, scrolling.as_deref_mut());
+    resume_touchpad_gesture(resumes_gesture, touchpad_down, scrolling.as_deref_mut());
 
     if touchpad_down && !has_scroll_event && scrolling.is_none() {
         insert_touchpad_begin_state(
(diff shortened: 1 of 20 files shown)


src/ecs/scroll.rs
```diff
@@ -26,7 +26,9 @@ pub struct ScrollEventsPlugin;
 
 const NATIVE_SCROLL_RESPONSE_SECONDS: f64 = 0.04;
 const NATIVE_SCROLL_SETTLE_PX: f64 = 0.25;
-const STICKY_EDGE_GAP_PX: i32 = 12;
+/// Distance from a window edge at which sticky scrolling engages. This is a
+/// hit zone, not a visual gap: the resulting snap lands exactly on the edge.
+const STICKY_EDGE_THRESHOLD_PX: i32 = 12;
```
karinushka: I think this should be a "screen width ratio". Hardcoding pixels is inflexible for various display resolutions.
Here's how I did it in my clone:
```rust
const STICKY_EDGE_THRESHOLD_RATIO: f64 = 0.1;
...
let threshold = (f64::from(viewport.width()) * STICKY_EDGE_THRESHOLD_RATIO) as i64;
```

karinushka:
I copied this into my local branch (adjusted the screen ratio snap to 0.1 or 10%).

There are some weird issues with it though - did not debug it too deep. When scrolling around the window sometimes snaps correctly, while other times it snaps as if the neighbor window has focus. So there is maybe an issue figuring out which window to actually snap - maybe the easiest is to track which window currently has the mouse in.
Also the momentum commit seems to continuously scroll without any decay.

Otherwise it's a cool feature - should be nice on smaller screens like laptops.

MrFlashAccount:
Addressed the review feedback in 5a469a3:

- paging now selects the managed window under the pointer for modifier-scroll, while raw multi-finger gestures remain position-based;
- the snap hit zone is now 10% of the usable viewport width;
- unsnapped origins cannot skip their first stop;
- native momentum start/end commits the interpolation target, so there is no extra synthetic decay tail after macOS momentum;
- the full test suite is green (188/188), and production Clippy passes with warnings denied.

I kept the PR as a draft only for hands-on feel validation on real displays.

karinushka:
So when changing to a different macOS workspace, the scroll momentum disappears permanently. E.g. you try to swipe and the windows move 1cm at a time.
Also I am testing on a strip with 3 windows. Snapping to left works, swipe + snap to middle works.. and then the last one which is semi-offscreen, the swipe refuses to move.
Manually clicking the third window brings it into view - and then the swipe works the other way around. Swipe + snap to the middle window, but then the first window is fully out of view and refuses to scroll further. Using `focus_east` on it gives it focus, but it's still hidden from the view.
I am testing on a large external monitor where 2-3 windows are visible at the same time, so maybe this causes some issues.

So there seems to be a bunch of edgecases broken - give it a spin and see whether you can convince the LLM to fix it 😺
Otherwise I can also dive deep this weekend and try to clean it up.

MrFlashAccount:
Thanks — I managed to reproduce the two underlying failures deterministically on the previous PR head (`5a469a3`), without relying on trackpad feel:

1. Activating a different native macOS workspace left `Scrolling` and `RepositionMarker` attached to the old strip. That stale state explains the permanently shortened movement after a workspace switch and could also overwrite the reposition requested by `focus_east`.
2. On an 1800-point viewport with three 1200-point columns, starting at `-900` with the semi-visible third window selected by the pointer chose `-1800` and skipped the adjacent `-1200` stop.

Both regressions already pass on the current fork `main`. I ported only the minimal pieces into this PR in `d52ddec` rather than cherry-picking `ff89a05` wholesale, since that commit also contains unrelated event-runtime and performance work:

- clear transient scroll/reposition state when leaving a native workspace;
- remove the pointer-selected paging anchor and derive the one-hop neighborhood from the actual strip position and nearest global stops.

Verification is green: 189/189 tests and Clippy with warnings denied. The exact `focus_east` sequence is not AX-UI-tested, but its existing explicit reshuffle can no longer be overridden by the stale workspace scroll state, and the focus integration suite remains green.

Could you retry the same large-monitor / three-window setup on the updated branch?

karinushka:
https://github.com/user-attachments/assets/5495b03c-bf49-4886-b0db-57b1bac58fb3

There is still an issue with registering focus. You can see in the recorded video (unfortunately Github allows only 10mb size) - scrolling once to the next window, and then it refuses to scroll on... until I click in it with the mouse. Then it scrolls to next window and again refuses to scroll.. until a mouse click.

Also, can you rebase this on top of `testing` again - some of the earlier commits from this PR are already there.

src/menubar.rs
karinushka, line 294: This runs as a regular Bevy system in `PostUpdate`, e.g. on every frame tick of the main Bevy loop, it executes queries, checks active display bounds, formats
  string labels, rebuilds/iterates menu items, and makes multiple FFI Objective-C calls (`item.setState(...)`, `item.setEnabled(...)`, `button.setTitle(...))`.

Use Bevy `Local<T>` for state management or preferably use Bevy `Changed<LayoutStrip>`, `Changed<Position>`, `Changed<FocusedMarker>` filters to ensure ObjC status item mutations are executed only when active workspace, focused window, or window widths actually change.

src/manager/windows.rs
karinushka, line 513: `update_frame()`, `set_ax_position()`, and `set_ax_size()` issue synchronous IPC calls to macOS WindowServer. Running up to 3 round-trips in a blocking loop during `PostUpdate` execution on the main thread can cause frame drops and visual window jitter during fast animations or live resizes.

Guard staging so it only triggers when `size.x` > `display_width`.
Also add strict exit conditions so non-responsive applications do not stall the main thread for 3 iteration rounds.

src/menubar.rs
karinushka, line 172: `focused_width_ratio` is calculated in `update_menu_bar` from the window's physical pixel width (`bounds.0.x / display_width`). When a window has horizontal padding configured (`horizontal_padding`), its physical pixel width is reduced (e.g., 480px instead of 500px on a 1000px display = 48%).

src/ecs/scroll/paging.rs
```diff
@@ -0,0 +1,381 @@
...
+    let Some((lower_bound, upper_bound)) = content_bounds(viewport, &columns) else {
+        return Vec::new();
+    };
+
+    let mut stops = columns
+        .into_iter()
+        .flat_map(|column| column_stops(viewport, column, lower_bound, upper_bound))
+        .collect::<Vec<_>>();
+    stops.sort_by(|left, right| right.total_cmp(left));
+    stops.dedup();
```
karinushka, line 102: `stops.dedup()` relies on exact float equality (`==`). Floating-point coordinate calculations can produce microscopic sub-pixel discrepancies (e.g., -600.0000000000001 vs -600.0).

Use `stops.dedup_by(|a, b| (a - b).abs() < STOP_ALIGNMENT_EPSILON)`.

src/ecs/scroll.rs
```diff
@@ -76,80 +103,291 @@ fn swipe_gesture(
...
+}
+
+fn read_gesture_input(
+    messages: &mut MessageReader<Event>,
+    config: &Config,
+    scroll_scale: f64,
+) -> GestureInput {
+    let mut input = GestureInput::default();
+    for event in messages.read() {
+        match event {
```
karinushka, line 254: `read_gesture_input` flattens all events received during a single frame into a u8 bitfield.
If TouchpadUp and a subsequent TouchpadDown arrive within the same frame tick, the temporal sequence is destroyed. In swipe_gesture, lifecycle methods execute in a static sequence (begin ➔ end ➔ resume ➔ finish), causing `finish_touchpad_gesture` (`TOUCHPAD_UP`) to run last and immediately cancel the newly begun touch gesture.

Process gesture lifecycle state transitions in true queue order from `messages.read()` rather than packing them into an unordered bitfield.

src/menubar.rs
karinushka: I fixed it in https://github.com/karinushka/paneru/commit/d70274ce9731835e718107be155fa4cd8741ae60
Should now only run on focus change.

PR #308, opened 2026-07-16
Event-driven runtime and per-strip scroll scheduling

MrFlashAccount:
> [!WARNING]
> **STACKED ON #307 — review after #307.**
> The previous mixed runtime + observation + persistence patch has been split. This PR now contains only the event-driven runtime and scroll scheduling slice.

## Why

Paneru used fixed-rate polling even while idle, and a single global scroll state could make independent strips interfere with one another. This patch makes the runtime wake for actual work or explicit deadlines and keeps scroll animation state per strip.

## What changed

- pump native events before all Bevy event readers
- wake the main run loop when background or synthetic events arrive
- replace repeated timer polling with monotonic deadlines
- preserve the existing one-minute low-power/state-save fallback wake
- keep upstream brute-force window discovery unchanged
- keep existing persistence and restore behavior unchanged
- scope scrolling, timeout cleanup, and reconciliation to each strip
- narrow the input event mask to configured features
- keep Accessibility/CoreGraphics calls that require it on the main thread
- make configuration reload replace input taps transactionally

## Explicitly not in this PR

- Accessibility window observers
- persistence coalescing or restore-policy changes
- opt-in window management
- updater or menu-bar behavior

Those changes will be proposed separately so each trade-off can be reviewed independently.

- #312 — scoped window observation and coalesced persistence
- #309 — optional per-window management
- #311 — display-synchronized frame pacing

## Verification

- `cargo test` — 208 passed
- regression coverage includes queued-event wakeups, synthetic-event latching, monotonic deadlines, per-strip scrolling, main-thread AX/CG access, and input-mask transitions

## Pre-merge validation

- repeat idle and active-scroll power sampling on the isolated branch
- manually verify native scroll feel on 60 Hz and ProMotion displays

The code is ready for review; these measurements remain required before merge.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -67,6 +67,8 @@ Configure trackpad gestures and scroll-wheel window sliding.
 | `direction` | String | `"Natural"` | Direction of movement: `"Natural"` or `"Reversed"`. |
 | `vertical` | Boolean | `true` | Interpret the vertical gestures with `fingers_count` or ignore them. Enabling this allows using vertical swipe gestures to change virtual desktops. |
 
+When `fingers_count` is omitted or set below 3, Paneru does not intercept native macOS gestures. If macOS uses three-finger horizontal swipes for Spaces, prefer `[swipe.scroll]` with a modifier or configure a different finger count.
+
 ### `[swipe.scroll]`
 | Option | Type | Default | Description |
 | :--- | :--- | :--- | :--- |
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -59,7 +59,7 @@ Configure trackpad gestures and scroll-wheel window sliding.
 | `sensitivity` | Float (0.1–2.0) | `0.35` | Multiplier for swipe distance. |
 | `deceleration` | Float (1.0–10.0) | `4.0` | Rate at which inertia slows down after a swipe. |
 | `continuous` | Boolean | `true` | Controls strip edge limits while scrolling; disable to clamp strictly to the full content bounds. |
-| `sticky` | Boolean | `false` | After scrolling ends, smoothly snap the nearest column to the center of the viewport. |
+| `sticky` | Boolean | `false` | After scrolling ends, smoothly snap the nearest column edge to the matching viewport edge, preserving a small outer gap. |
 
 ### `[swipe.gesture]`
 | Option | Type | Default | Description |
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -59,7 +59,8 @@ Configure trackpad gestures and scroll-wheel window sliding.
 | `sensitivity` | Float (0.1–2.0) | `0.35` | Multiplier for swipe distance. |
 | `deceleration` | Float (1.0–10.0) | `4.0` | Rate at which inertia slows down after a swipe. |
 | `continuous` | Boolean | `true` | Controls strip edge limits while scrolling; disable to clamp strictly to the full content bounds. |
-| `sticky` | Boolean | `false` | After scrolling ends, smoothly snap the nearest column edge to the matching viewport edge, preserving a small outer gap. |
+| `paging` | Boolean | `true` | Limit one gesture to at most one adjacent stop. A regular column has one stop; an oversized column has exactly two, at its left and right edges. Set to `false` to restore free scrolling. |
+| `sticky` | Boolean | `false` | With `paging = false`, snap exactly to a column edge when scrolling ends inside its 32-point hit zone. This activation zone does not create a visual gap. |
 
 ### `[swipe.gesture]`
 | Option | Type | Default | Description |
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -60,7 +60,7 @@ Configure trackpad gestures and scroll-wheel window sliding.
 | `deceleration` | Float (1.0–10.0) | `4.0` | Rate at which inertia slows down after a swipe. |
 | `continuous` | Boolean | `true` | Controls strip edge limits while scrolling; disable to clamp strictly to the full content bounds. |
 | `paging` | Boolean | `true` | Limit one gesture to at most one adjacent stop. A regular column has one stop; an oversized column has exactly two, at its left and right edges. Set to `false` to restore free scrolling. |
-| `sticky` | Boolean | `false` | With `paging = false`, snap exactly to a column edge when scrolling ends inside its 32-point hit zone. This activation zone does not create a visual gap. |
+| `sticky` | Boolean | `false` | With `paging = false`, snap exactly to a column edge when scrolling ends inside a hit zone equal to 10% of the usable viewport width. This activation zone does not create a visual gap. |
 
 ### `[swipe.gesture]`
 | Option | Type | Default | Description |
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -119,6 +119,7 @@ Format: `"[modifiers-]key"`. Available modifiers are:
 - `ctrl`, `lctrl`, `rctrl`
 - `cmd`, `lcmd`, `rcmd`
 - `shift`, `lshift`, `rshift`
+- `fn`
 
 | Action | Description |
 | :--- | :--- |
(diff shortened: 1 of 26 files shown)


MrFlashAccount:
Rebuilt and force-pushed as a runtime-only slice. Window observation and persistence moved to #312, optional per-window management remains in #309, and display pacing moved to #311. This head preserves upstream brute-force discovery and existing persistence behavior. Verification: 208 tests passed; strict production clippy passed.

karinushka:
Have you tried running this PR and do you notice any significant performance increases, such as smoother scrolling or what not? Potentially try to do some measurements - on whether there is an actual performance benefit in switching the model.

I am curious, because @georgiansarghi did a bunch of performance optimizations in their fork (https://github.com/georgiansarghi/paneru/pull/2), which on paper looked promising, but when measured in practice did not result in anything significant.
Unfortunately I can not find the summary in that PR - maybe it disappeared when the branch was deleted. Ping to @georgiansarghi to drop some knowledge into this thread 😺

Bevy has parallel iterators which I used at some point in multiple systems (animations, window resizing)... but then I realized that scheduling these multithreaded loops takes some overhead, so it is only worth when you have thousands of objects to operate on - not 5 strips and 15 windows. (I.e. is it worth to have per-strip scroll scheduling when there are only a few strips present and at most one of them is scrolling at a time)

georgiansarghi:
I did see better idle performance with my branch, but I don’t have proper numbers to back it up. I think #308 is a cleaner version of the same general idea.

The main reason I made that branch was idle CPU usage. Paneru  was constantly using around 3–5% CPU on my machine while doing nothing, sometimes more. With my branch, idle CPU was noticeably lower. I didn’t do a proper benchmark, and unfortunately I no longer have the measurements so please treat this as a personal observation.

The main change was replacing the loop that woke up roughly every 50 ms with one that could sleep for longer. Events would wake it immediately, so commands and input should still respond quickly. I also experimented with running Bevy systems on one thread and checking native tabs less often. I’m not sure how much those parts helped compared with simply waking the loop less often.

I didn’t propose it upstream because the custom runner was fairly complicated. It had to track timers, wakeups, internal Bevy work, and a few fallback cases. The upstream model was simpler and probably more robust.

Looking at #308, it seems to use the same basic idea but keeps more of Bevy’s normal runtime model. On paper, it should be able to sleep even longer than my branch while fully idle, so I think it is probably the better direction for upstream.

Also, the per-strip scrolling code in #308 doesn’t appear to start separate parallel jobs for each strip. It mostly loops over the strips normally. I see that part more as a correctness/isolation change than a performance optimization. I wouldn’t expect it to make scrolling faster.

I also wouldn’t expect the event-driven runtime by itself to make active scrolling much smoother.

useful comparison would be:
- idle CPU time and wakeups;
- energy impact over a few minutes;
- command/input latency after being idle for 30–60 seconds.

karinushka:
Thank you for the explanation!
I think the simplest way to improve on the idle CPU usage is to increase the macOS pump event timeout:
https://github.com/karinushka/paneru/blob/e5cf00809a233d5ec17ad2f7ea9c8866a014cdc3/src/ecs/systems.rs#L582

If it's set to something quite high, say1sec - it will provide sufficient power throttling if no events are happening. As soon as something happens it will return right away, so there should be no lag.
Another potential improvement, now that I'm thinking of it, is to use Bevy's `FixedUpdate` schedule for the `pump_events` system. This will call it on regular intervals and not "per frame" together with other systems.

**Update**: yeah, those won't work. A higher timeout causes the cocoa pump loop to wait, and `FixedUpdate` runs on a quite slow schedule, so it causes huge lag.
So the easiest could be to bump the 50ms delay to something higher:
https://github.com/karinushka/paneru/blob/e5cf00809a233d5ec17ad2f7ea9c8866a014cdc3/src/ecs/systems.rs#L44

It will only result in the "initial" lag when the user comes back to the machine - then it immediately snap to the lower threshold again.

MrFlashAccount:
@karinushka I reran the measurements on the same Mac. The result supports the event-driven runtime primarily as an **idle power/CPU improvement**, not as an active-scrolling speedup.

Important scope caveat: this is not a perfectly isolated benchmark of #308. The old run used one managed 200%-wide TextEdit window; the new run used two managed windows (Arc + ChatGPT) on fork v0.6.6, which also contains the separate display-pacing work now split into #311. Therefore the idle comparison is useful, while the active-scroll peak cannot be attributed to #308 alone.

### Old fixed-rate/upstream-like runtime

- Steady idle: **0.9–1.0% CPU** (about 0.95%), 24 MB, 20 threads.
- Short-scroll peak: **13.2% CPU**.
- More than 100 seconds after scrolling: still **2.48% CPU**.
- Idle wakeups increased from **0.26/s before the gesture to 1.27/s afterward** (about **4.9x**), reproducing the persistent post-scroll wakeup tail.

### Event-driven fork runtime

- Settled idle: below `top`'s **0.1% CPU reporting floor**; p50 and p95 were both 0%, at 24–25 MB.
- During a settled-idle observation window, the idle-wakeup and context-switch counters did not increase.
- In a 150-second trace containing one gesture: mean **2.019%**, p95 **13.1%**, peak **26.8%**. The gesture window averaged **9.875%** over 16 seconds.
- The process returned to 0% roughly 60 seconds after the gesture and remained there for the final 75 seconds.
- There was also a delayed 19-second burst averaging about **5.99% CPU**; I have not explained that yet, so it remains a performance target.

### Calculations

- Conservative settled-idle reduction: `1 - 0.1 / 0.95 > 89%`. This is a lower bound because 0.1% is the measurement floor, not the measured mean.
- Post-scroll-tail reduction after settling: `1 - 0.1 / 2.48 > 96%`.
- Active peak: `26.8 / 13.2 = 2.03x` higher, so this change should **not** be presented as reducing active-scroll CPU.

The higher active peak is consistent with the separate display-pacing change: in the full fork, geometry application is synchronized with the display refresh cadence (up to 120 Hz on this ProMotion display) instead of a fixed 60 Hz timer. `120 / 60 = 2x`, close to the observed `2.03x` peak ratio. A locked 60 Hz vs 120 Hz A/B would still be required to prove that the entire difference comes from refresh rate. Once animation work settles, the display link stops and the process goes back to sleep.

I also agree with @georgiansarghi about the per-strip scheduler: it does not create parallel jobs and I do not consider it a scrolling performance optimization. Its purpose is correctness/isolation between independently managed strips.

So the concrete benefit measured so far is removal of the persistent ~1% idle polling cost and the post-scroll wakeup leak. I have not yet produced a direct Energy Impact measurement or instrumented command latency after 30–60 seconds of idle, so I would not claim those results yet.

MrFlashAccount:
A follow-up after rerunning the measurements: the scrolling improvement is substantial, not marginal.

### Current fork result

The latest Paneru 0.6.8 run used two managed windows and display-synchronized pacing at 120 Hz. Under those conditions:

- gesture phase: 2.43% average CPU
- active-scroll peak: 6.0% CPU
- immediate post-scroll phase: 0.25% average CPU
- after 100+ seconds: 0.24% average CPU
- delayed bursts peaked at only 2.0% and 1.5%

Scrolling also feels visibly smoother in actual use.

### Improvement over the previous fork run

The earlier v0.6.6 run also used two managed windows and 120 Hz pacing:

- gesture-window average: 9.875% to 2.43%, approximately 75.4% lower
- active-scroll peak: 26.8% to 6.0%, approximately 77.6% lower
- previous delayed burst: 5.99% average for 19 seconds, with samples up to 16%; the latest delayed peaks were only 2.0% and 1.5%

### Improvement over the older upstream-like baseline

The older baseline was an easier workload: one managed 200%-wide TextEdit window with fixed 60 Hz pacing. It still used more CPU:

- active-scroll peak: 13.2% to 6.0%, approximately 54.5% lower
- CPU after 100+ seconds: 2.48% to 0.24%, approximately 90.3% lower

So the current implementation is processing two managed windows at twice the display cadence, while producing smoother scrolling and using materially less CPU than the older one-window 60 Hz baseline. A rough peak-per-refresh normalization is 13.2 / 60 = 0.22 versus 6.0 / 120 = 0.05, around a 4.4x improvement. That normalization is only directional, but the raw result already shows the improvement without relying on it.

The runs are not a laboratory-perfect frame-by-frame A/B, but the current workload is harder, not easier, and the performance margin is large. I am therefore confident that the scrolling path in the current fork has been significantly optimized.

Important scope note: I have not ported or cherry-picked these newer scrolling optimizations into PR #308 yet. This PR still contains the earlier runtime-only slice, so these results must not be attributed to the current PR branch. I will keep the optimized scrolling changes separate until they can be isolated cleanly.

PR #309, opened 2026-07-16
[WIP / STACKED / UPSTREAM DECISION] Optional per-window management

MrFlashAccount:
> [!CAUTION]
> **WIP / STACKED — upstream product decision requested.**
> The implementation blockers are resolved, but this mode may still be better kept fork-only.

## Intent

Add an optional mode in which Paneru manages individual windows only after an explicit opt-in.

```toml
[options]
opt_in_management = true
```

The option defaults to `false`, so current upstream automatic-management behavior is unchanged.

## Behavior

- `opt_in_management = false`: ordinary windows remain managed by default
- `opt_in_management = true`: ordinary windows start as passthrough
- the keybinding, menu action, and IPC command target one concrete window
- toggling on captures the ordinary macOS frame; toggling off restores it
- passthrough windows remain focusable but are excluded from layout, scrolling, borders, dimming, geometry writes, and persistence
- inactive-Space windows are still discovered through the bounded brute-force path in both modes

## Deterministic ownership precedence

1. explicit `manage = false`
2. explicit `floating = true`
3. explicit `manage = true`
4. saved managed membership from session restore
5. configured default (`Managed` normally, `Passthrough` in opt-in mode)

Manual opt-ins are persisted through saved strip membership and restored on the next launch. Explicit passthrough or floating rules always override saved state.

## Dependencies

- #308 — event-driven runtime
- #312 — scoped observation and coalesced persistence

The branch is cumulative because GitHub cannot use a fork branch as the upstream base.

## Verification

- `cargo check --all-targets` — passed
- `cargo test` — 243 passed
- tests cover legacy default behavior, rule precedence, targeted multi-window toggles, frame restoration, lifecycle transitions, persistence isolation, saved manual opt-in restore, and explicit-rule override

## Upstream decision requested

Do you want this optional mode in upstream at all?

If not, #308 and #312 are independently useful performance/reliability changes, and this PR can remain fork-only.


MrFlashAccount:
Before I refine this branch further: do you want per-window opt-in management in upstream at all?

The current draft now preserves Paneru's existing behavior by default:

```toml
[options]
opt_in_management = false
```

Setting it to `true` makes ordinary windows passthrough until the user explicitly manages them through the keybinding, menu-bar action, or a matching `manage = true` rule.

Both modes retain the inactive-Space brute-force discovery path, so saved managed windows can still be found and restored before the user visits every Space.

If this mode does not fit the upstream direction, I can keep #309 fork-only and avoid adding its ownership/persistence complexity to the upstream series. Which direction would you prefer?

MrFlashAccount:
Rebuilt on top of the split runtime/foundation branches. The implementation blockers are resolved: legacy automatic management remains the default, inactive-Space brute-force discovery is preserved in both modes, and ownership precedence is now deterministic and tested. Saved manual opt-ins restore across launches, while explicit manage=false or floating=true rules win. The remaining blocker is the upstream product decision: should this optional mode be upstream at all?

PR #310, opened 2026-07-16
Accessibility permission onboarding without prompt loops

MrFlashAccount:
> [!WARNING]
> **Manual macOS TCC validation remains before merge.**
> This PR is now Accessibility-only. Display frame pacing has been removed and will be proposed separately on top of the runtime PR.

## Problem

`AXIsProcessTrustedWithOptions(... prompt = true)` was used as a permission check. Repeating that check could repeatedly ask macOS to show the Accessibility prompt.

## What changed

- split the silent trust check from the one-shot system prompt request
- show a native setup explanation before requesting the system prompt
- explain how to remove and re-add a stale Paneru entry in System Settings
- keep Paneru alive in the menu bar while permission is missing
- provide menu actions to reopen the instructions or Accessibility settings
- start the window manager automatically after permission becomes available
- allow quitting cleanly while Paneru is waiting

The passive wait loop uses `AXIsProcessTrusted()` only. The prompting API is called only after an explicit **Continue** action.

## Independence

- based directly on `testing`
- no dependency on opt-in management
- no dependency on the event-driven runtime
- no display-link or animation changes

## Verification

- `cargo check --all-targets` — passed
- `cargo test` — 161 passed

## Pre-merge manual validation

- first launch with no Accessibility entry
- denied / **Not Now** flow
- stale already-listed TCC entry
- granting permission while Paneru is waiting
- quitting from the permission-required menu

Those cases modify real macOS TCC state. The code is ready for review, but they must be exercised on a packaged app before merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ notify = { version = "8.2", features = ["macos_fsevent"] }
 num_enum = "0.7"
 objc2 = { version = "0.6", features = ["catch-all", "exception"] }
 objc2-app-kit = { version = "0.3", features = [
+  "NSAlert",
   "NSBezierPath",
   "NSColor",
   "NSGraphics",
--- a/src/accessibility_prompt.rs
+++ b/src/accessibility_prompt.rs
@@ -1,0 +1,38 @@
+use objc2::MainThreadMarker;
+use objc2_app_kit::{NSAlert, NSAlertFirstButtonReturn, NSApplication};
+use objc2_foundation::NSString;
+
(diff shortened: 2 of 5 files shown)


MrFlashAccount:
The requested split is complete. This PR is now Accessibility-only and based directly on testing; display-synchronized pacing is in #311. The passive permission loop never invokes the prompting API, and the system prompt is requested only after an explicit Continue action. Verification: 161 tests passed; strict production clippy passed.

karinushka:
This one is super cool for new users who onboard!

PR #327, opened 2026-07-24
fix: Expose moved window and refresh menu bar on follow-move to virtual workspace

skmono:
## Problem

When moving a window to another virtual workspace with focus-follow
(`virtualmove` / `virtualmovenum`), two things were broken:

1. **The moved window could end up off-screen.** After the move, the window was
   appended to the far right of the destination strip and stayed there,
   partially or fully off the right edge until it was manually centered or
   re-focused.

2. **The menu bar's virtual workspace label did not update** to reflect the new
   active VW.

## Root cause

Both stem from the fact that on a follow-move the moved window *keeps* focus, so
nothing that keys off `Added<FocusedMarker>` fires:

- `handle_virtual_window_moves` appends the window, activates the destination
  workspace, and calls `reshuffle_around`, all in one frame. But
  `reshuffle_layout_strip` removes the reshuffle marker and then skips the
  strip because its `ActiveWorkspaceMarker` is `is_added()` that frame (a
  deliberate guard against a workspace-switch scheduling race). The reshuffle
  request never occurs, and `show_active_workspace` only restores a saved
  scroll position, which a freshly-appended window isn't part of.

- `update_menu_bar` ran only on `Added<FocusedMarker>`hence the focused window is
  unchanged on a follow-move, so it never re-ran.

## Fix

- **Visibility:** In the append branch of `handle_virtual_window_moves`, force a
  genuine re-focus of the moved window (remove `FocusedMarker` first, since
  Bevy's `try_insert` on a present component won't re-fire `Added`). This makes
  `autocenter_window_on_focus` re-issue the reshuffle on a later frame after the
  active marker is no longer `is_added`, bringing the window on screen.
  Gated on `mid_placement.is_none()` so the `insert_windows_mid_strip` feature's
  "keep exact x / no animation" invariant is untouched (that path already keeps
  the window visible).

- **Menu bar:** `update_menu_bar` now also refreshes on
  `Added<ActiveWorkspaceMarker>`, so the VW label updates on any active-workspace
  change, covering both append and mid-strip follow-moves, independent of focus.

## Testing

- Added `test_follow_move_brings_appended_window_on_screen`, which seeds a
  destination strip wider than the display and asserts an appended follow-moved
  window is brought fully on-screen. Verified it fails without the fix (window at
  x 800..1200 on a 1024px display) and passes with it.
- Full suite: 173 passed, clippy clean.

### DIFF
--- a/src/menubar.rs
+++ b/src/menubar.rs
@@ -1,6 +1,5 @@
-use bevy::ecs::entity::Entity;
 use bevy::ecs::query::{Added, Has, With};
-use bevy::ecs::system::{NonSendMut, Populated, Query, Res};
+use bevy::ecs::system::{NonSendMut, Query, Res};
 use objc2::rc::Retained;
 use objc2::{DefinedClass, MainThreadMarker, MainThreadOnly, define_class, msg_send, sel};
 use objc2_app_kit::{
@@ -15,7 +14,7 @@ use crate::accessibility_prompt::{AccessibilitySetupAction, show_accessibility_s
 use crate::commands::{Command, Operation};
 use crate::config::Config;
 use crate::ecs::params::ActiveDisplay;
-use crate::ecs::{Bounds, FocusedMarker, Unmanaged};
+use crate::ecs::{ActiveWorkspaceMarker, Bounds, FocusedMarker, Unmanaged};
 use crate::events::{Event, EventSender};
 use crate::manager::request_ax_privilege;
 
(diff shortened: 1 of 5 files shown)


src/menubar.rs
```diff
@@ -300,12 +299,16 @@ impl Drop for MenuBarManager {
 
 #[allow(clippy::needless_pass_by_value, clippy::type_complexity)]
 pub fn update_menu_bar(
-    _guard: Populated<Entity, Added<FocusedMarker>>,
+    focus_changed: Query<(), Added<FocusedMarker>>,
+    workspace_changed: Query<(), Added<ActiveWorkspaceMarker>>,
```
karinushka: Use `Populated` - that way the system won't even schedule if there are no changes. Then there is no need to do a query check below.
You can even add both conditions together with `Or<>`, something like:
`Populated<Or<Added<FocusedMarker>, Added<ActiveWorkspaceMarker>>>`

(For some reason LLMs are not aware of `Populated` function 🤷)

karinushka (commented):
Can you rebase this to `testing` branch - that way we have a chance to test it before unleashing it on the main population.

skmono:
> Can you rebase this to `testing` branch - that way we have a chance to test it before unleashing it on the main population.

Rebased and reverted back to use `Populated`.

karinushka:
I rebased this onto testing with you as the author: https://github.com/karinushka/paneru/commit/1d25989f7affa1ec0cd47a8969ab31656d3e3498
Check it out if it works as expected.

skmono:
> I rebased this onto testing with you as the author: [1d25989](https://github.com/karinushka/paneru/commit/1d25989f7affa1ec0cd47a8969ab31656d3e3498) Check it out if it works as expected.

Tested and verified everything working as intended

PR #336, opened 2026-07-29
fix: clamp windows after display disconnect

### LINKED ISSUE #332: Support for enforcing maximum window widths
When I disconnect my monitor, which is wider than my built-in display, my windows stay wider than the built-in display. It would be nice if the windows that were wider than the screen got resized to fullscreen's size.

Not sure if this is intentional or not.

hfl0506:
## Related Issue
Closes #332 

## Changes
- Resize oversized windows after disconnecting a larger external display.
- Clamp managed-window widths to the replacement display’s usable width.
- Clamp floating-window width and height independently without enlarging dimensions that
  already fit.

- Move floating windows into the replacement display’s usable area.
- Retry clamping after transient workspace-enumeration or AX frame-read failures.
- Continue processing unrelated windows when one window fails.
- Add end-to-end coverage for display removal, floating geometry, and transient failures.
- Extend test mocks with controllable workspace-query and frame-read failures.
- Pass formatting, cargo check, all 174 tests, and strict Clippy.

## Test Result
```bash
cargo test test_unplug_clamps_windows_to_the_replacement_display_bounds -- --nocapture
```

### DIFF
--- a/src/ecs/display.rs
+++ b/src/ecs/display.rs
@@ -31,6 +31,9 @@ const ORPHANED_SPACES_TIMEOUT_SEC: u64 = 30;
 
 pub struct DisplayEventsPlugin;
 
+#[derive(Component)]
+pub(crate) struct ClampWindowBounds;
+
 impl Plugin for DisplayEventsPlugin {
     fn build(&self, app: &mut App) {
         app.add_systems(PreUpdate, display_change_handler);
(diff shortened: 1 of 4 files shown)


karinushka:
When reviewing this PR I realized that it is actually somewhat simpler to do. Check it out in `testing` branch: https://github.com/karinushka/paneru/commit/f72b0271e5310ceee9d2701c8b8736f04f09037f

There is already a marker (`RefreshWindowSizes`) which is inserted in all workspaces which have been moved during display changes - and that marker waits for 5 seconds and then triggers.
Once that happens the fix is to just run through the windows and correctly resize them according to their parent display.
The issue in the original code was that it was only working on active display - so some windows could have gone unnoticed.

hfl0506:
hi @karinushka okay, let's use the simple implementation instead. Can I close this MR accordingly ? Thanks

PR #337, opened 2026-07-31
Add lua module

auscyber:
Disclaimer: i did use claude code for a lot of things, but I do have a lot of experience in rust and with mlua as well and changed this to be effective

So this adds a lua config capability and in order to facilitate that i made the commands and query state in a separate crate with strict types so you can determine coherence between the cli and the lua api. that is in a separate commit

i also wanted to add querying whats on screen for my own personal uses, and i wasnt able to split it into a separate thing

crane is ideal for this so dependencies can be built only once and then shared between the two outputs

i also had to make sdk_dir properly use the shell developer dir otherwise it doesnt build as nocc is not correct for lua environments, 

i am willing to change a lot though and see how it goes, at the moment im testing and using this myself

### DIFF
--- a/crates/lua/src/client.rs
+++ b/crates/lua/src/client.rs
@@ -1,21 +1,11 @@
-//! The client half of the API: talking to a running daemon over its Mach
-//! service.
+//! The client half of the API: talks to a running daemon over its Mach
+//! service. [`module`] builds the table `require("paneru")` returns.
 //!
-//! [`module`] builds the complete table that `require("paneru")` returns — the
-//! shared typed API from [`crate::install`], dispatching to the daemon, plus the
-//! client-only `query_*` / `subscribe` / service-name helpers. The loadable
-//! `module` feature's `luaopen_paneru` is then a one-liner returning this table.
+//! Service name defaults to `com.karinushka.paneru`, overridable via the
+//! `PANERU_MACH_SERVICE` environment variable or `paneru.set_service_name`.
 //!
-//! The service name defaults to `com.karinushka.paneru`, can be overridden with
-//! the `PANERU_MACH_SERVICE` environment variable, or set at runtime via
-//! `paneru.set_service_name("com.example.paneru")`.
-//!
-//! Every call here is a blocking round trip, because Lua's C API is synchronous
-//! and a callback cannot yield into an executor. That is what
-//! [`paneru_mach_ipc`]'s `*_blocking` half is for: the wait is one kernel wait,
-//! where driving the async half with a `block_on` would arm a port watcher,
-//! build a waker and park the thread through an executor only to end up blocked
-//! on the same message.
+//! Every call here blocks: Lua's C API is synchronous, so a callback cannot
+//! yield into an executor.
 
 use std::rc::Rc;
 use std::sync::{LazyLock, Mutex};
@@ -31,8 +21,8 @@ use paneru_shared_types::wire::{
     Request, Response, ScriptStateRequest, ScriptStateResponse, WriteOutcome,
 };
 
-/// The active service name, seeded from the shared default (and its environment
-/// override) and mutable via `set_service_name`.
+/// The active service name, seeded from the shared default and mutable via
+/// `set_service_name`.
 static SERVICE: LazyLock<Mutex<String>> =
     LazyLock::new(|| Mutex::new(paneru_shared_types::wire::service_name()));
 
@@ -43,7 +33,6 @@ fn service_name() -> String {
     )
 }
 
-/// Connects to the running daemon.
 fn connect() -> LuaResult<Sender<Request>> {
     Sender::connect(&service_name()).map_err(|err| match err {
         paneru_mach_ipc::Error::NotRunning => {
@@ -53,11 +42,9 @@ fn connect() -> LuaResult<Sender<Request>> {
     })
 }
 
-/// Sends a request and waits for the answer.
-///
-/// The single place this module blocks. A daemon that reports a failure raises
-/// it as a Lua error rather than returning it, so a script sees a failed call as
-/// a failure instead of as a silent no-op.
+/// Sends a request and waits for the answer. A daemon-reported failure is
+/// raised as a Lua error rather than returned, so a failed call is an error,
+/// not a silent no-op.
 fn call(request: &Request) -> LuaResult<Response> {
     let sender = connect()?;
     let response: Response = sender.call_blocking(request).map_err(LuaError::external)?;
@@ -68,22 +55,16 @@ fn call(request: &Request) -> LuaResult<Response> {
     }
 }
 
-/// Sends a request that expects no answer.
 fn send(request: &Request) -> LuaResult<()> {
     let sender = connect()?;
     sender.send_blocking(request).map_err(LuaError::external)
 }
 
-/// The primitive the shared API is built on: send the command to the daemon
-/// (fire-and-forget).
-// Takes `Command` by value to match the shared `crate::Dispatch` closure type,
-// which every verb closure in `lib.rs` also has to satisfy.
 fn dispatch(_: &Lua, command: Command) -> LuaResult<bool> {
     send(&Request::Command(command))?;
     Ok(true)
 }
 
-/// Runs a state query and returns the payload.
 fn query_payload(kind: StateQueryKind) -> LuaResult<paneru_shared_types::wire::QueryPayload> {
     match call(&Request::Query(kind))? {
         Response::Query(payload) => Ok(payload),
@@ -91,7 +72,6 @@ fn query_payload(kind: StateQueryKind) -> LuaResult<paneru_shared_types::wire::Q
     }
 }
 
-/// Reads the `kind` argument shared by `query` and `query_json`.
 fn read_kind(kind: Option<String>) -> LuaResult<StateQueryKind> {
     let token = kind.unwrap_or_else(|| "state".to_string());
     StateQueryKind::parse(&token).ok_or_else(|| {
@@ -103,9 +83,6 @@ fn read_kind(kind: Option<String>) -> LuaResult<StateQueryKind> {
 }
 
 /// `paneru.query(kind)` — run a state query and return the raw JSON string.
-///
-/// Kept for scripts that want to hand the text to something else; `query_json`
-/// is what a script reading the answer itself wants.
 fn query(_: &Lua, kind: Option<String>) -> LuaResult<String> {
     Ok(query_payload(read_kind(kind)?)?
         .to_json()
@@ -121,7 +98,6 @@ fn query_json(lua: &Lua, kind: Option<String>) -> LuaResult<LuaValue> {
     lua.to_value(&json)
 }
 
-/// Runs one script-state request.
 fn script_state(request: ScriptStateRequest) -> LuaResult<ScriptStateResponse> {
     match call(&Request::ScriptState(request))? {
         Response::ScriptState(answer) => Ok(answer),
@@ -129,10 +105,8 @@ fn script_state(request: ScriptStateRequest) -> LuaResult<ScriptStateResponse> {
     }
 }
 
-/// Reads the current value of `key`.
-///
-/// A stored `Null` and an absent key both read as `nil` in Lua, which is the
-/// only thing Lua can express — it has no way to hold "present but empty".
+/// A stored `Null` and an absent key both read as `nil` in Lua — it has no way
+/// to hold "present but empty".
 fn state_get(key: &str) -> LuaResult<Option<ScriptValue>> {
     match script_state(ScriptStateRequest::Get {
         key: key.to_string(),
@@ -144,7 +118,6 @@ fn state_get(key: &str) -> LuaResult<Option<ScriptValue>> {
     }
 }
 
-/// Runs a write and returns its outcome.
 fn state_write(write: ScriptStateWrite) -> LuaResult<WriteOutcome> {
     match script_state(ScriptStateRequest::Write(write))? {
         ScriptStateResponse::Write(outcome) => Ok(outcome),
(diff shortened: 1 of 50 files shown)


auscyber:
btw, you can like not merge and entirely ignore this im adding a few nice to haves for me personally, cause i want to use this with sketchybar and have it call sketchybar lua things which i think is a natural implementation, but i can split this up if need be, make changes etc etc, i

 will likely continue using this myself as my own implementation i guess, but likely going to be obviously useful to other people as its a goal stated in the readme. but i do not want to be the type of person to throw slop and insane workloads at people,
 
 also to vouch that im like a real competent dev, ive used mlua before in wezterm https://github.com/wezterm/wezterm/pull/989,  im a nixpkgs contributor, the code was reviewed by me

karinushka:
Thanks! I patched this in locally and checking it out. To make it a bit easier, rebase this onto `testing` branch - that way we can test it easier without disturbing the `main` population.

Edit: Looks really cool! I especially like the ability to bind Lua functions to key bindings.

src/lua.rs
karinushka, line 1: Change this to new mod style, e.g. no `mod.rs` - then it matches with the rest of Paneru:
- src/lua.rs
- src/lua/api.rs
- src/lua/convert.rs

auscyber:
done!

auscyber:
i’m trying to clean it up a lot more manually as well. unfortunately  it’s config toml and lua but it would be exceptionally nice to make it able to configure everything in lua with a set api but that’s a lot more confusing. this more allows interfacing and event management


https://github.com/auscyber/dotfiles/blob/master/aspects/wms/paneru/default.nix

currently how i’m using it

auscyber:
okay also added paneru.setup, im going to test it and stress test it

one problem currently having is that i think the lua runtime should probbly be in a separate thread somehow with.message passing perhaps, as running the lua code may block the main thread which isnt particularly ideal and can cause slow performance

auscyber:
i also would like love love love to add window management programatically in a way like xmonad queries, im a haskell and rust programmer mainly so my aim is to have it like that

auscyber:
okay idk how it ballooned to 10 thousand lines, its a pretty reasonable implementation

karinushka:
> one problem currently having is that i think the lua runtime should probbly be in a separate thread somehow with.message passing perhaps, as running the lua code may block the main thread which isnt particularly ideal and can cause slow performance

Yeah, the standard way to integrate that stuff with Bevy data model is to spawn a background thread and run the Lua runtime there. Bevy has a compute thread pool for that:
```rust
let thread_pool = bevy::tasks::AsyncComputeTaskPool::get();
thread_pool.spawn(async move {
...
}
```
Does Lua runtime support running as `async`?

Due to the way Bevy multiplexes components in its systems, the only way for that background thread to communicate is by having a message bus. Similar to how Paneru handles the OS layer callbacks - they are sent upwards through a MPSC channel... and then distributed using Bevy's native messages.

src/lua/worker.rs
karinushka, line 1: Ah, so now I reached your thread commit 😺

Yeah, I am not sure what to pick - a Bevy native compute thread pool or just leave it as an OS thread like this.
On one hand having OS threads keeps this module clean of Bevy dependencies.. while on the other hand using a Bevy thread pool feels neater and also probably comes with built-in cleanup etc. 🤷

auscyber:
i see i see, probably better to use bevy threads, cause the performance without it is pretty horrendous.  but i also doubt itll be possible with bevy threads because the runtime doesnt have Send

Thanks
Ivy
________________________________

CONFIGURATION.md
```diff
@@ -456,6 +456,130 @@ the window manager on its next scheduled pass rather than instantly: expect the
...
+
+```lua
+local STASH = 9  -- a workspace you never switch to
+
+local pads = {
+  terminal = { match = paneru.match{ app = "Alacritty", title = "^scratch" },
+               spawn = "open -na Alacritty --args --title scratch" },
+  notes    = { match = paneru.match{ app = "Obsidian" },
+               spawn = "open -a Obsidian" },
+}
```
karinushka: Amazing Sauce!
`<meme>Vince McMahon falling over in amazement</meme>`

auscyber: I KNOW I KNOW, this was like a big big win for me, and a big motivation. being able to do thing slike this  was what i wanted. im still figuring out performance though as it can be quite mid

src/lua/worker.rs
auscyber: it has to be on os thread unfortunately, because lua is !Send but thats fine, im having a look at splitting the handlers into like concurrent sub threads, because it can cause updates to be slow if you have a lot of handlers. im gonna kinda force luajit for this tho, because its really really annoying to have it this slow.

CONFIGURATION.md
auscyber: this code doesnt work though and i have my own implementationthat ill move into here, but i see this as being truly the way to avoid bike shedding on features if people can just add the themselves in lua

auscyber:
also someone may say why lua bla bla bla why this structure, do you agree luajit especially is the most sane and most obvious way to do config like this?

auscyber:
and here is the scratchpad implementation i have 
https://github.com/auscyber/dotfiles/blob/master/aspects/wms/paneru/scratchpad.lua
pretty cool i'd say

auscyber:
its massively growing in scope, but im getting massive performance penalties of the code claude has made is not very idempotent, and avoids using async, im rewriting large parts of it to focus more on asynchronous and noblocking so the ax elements dont lock up as currently it is and it freezes every minute or so

auscyber:
claude seems to be allergic to async rust which im redoing so it actually does, it also loves to duplicate code so im cleaning that up where i can as well

karinushka:
If possible, can you just rebase-stack the two branches you have (lua-dedup and lua-stuff) instead of merging them. This makes the individual commits easier to follow. Also for some reason github reports a conflict - maybe it got confused by the merge 😺

I copied them from your fork and untangled them in my `jj` workspace when I was browsing the code - moving branches is very easy in jj. So if merge conflicts bog you down, I can push my copy - it will have you as author, but not have your signature (verified label on github).

auscyber:
> If possible, can you just rebase-stack the two branches you have (lua-dedup and lua-stuff) instead of merging them. This makes the individual commits easier to follow. Also for some reason github reports a conflict - maybe it got confused by the merge 😺
> 
> I copied them from your fork and untangled them in my `jj` workspace when I was browsing the code - moving branches is very easy in jj. So if merge conflicts bog you down, I can push my copy - it will have you as author, but not have your signature (verified label on github).

oh yeah im in jj too, i should just un track one of them hang on actually i dont know how to rebase stack. like what do i do?

karinushka:
> should just un track one of them

So the chain starts in:
`refactor: extract shared types into shared_types crate`
... runs until
`refactor(lua): share dispatch tail and skip idle event marshalling`
and here it splits into two directions:
* a long chain from `refactor(lua): split convert into world-side data and Lua-side marshalling` to `add some example scratchpad stuff`
* a chain with two entries: `feat(lua): configure everything from init.lua via paneru.setup` and `feat(nix): sketchybar-style `config` (init.lua) option; share module code`.

Instead of merging these two chains in one commit `merge add-lua-stuff into lua-dedup`, put the shorter one (with 2 commits) on top of the long one:
```
jj rebase -s [short start] -d [long end]
```

auscyber:
okay this should be right

karinushka:
Your last commit with "close layout gaps..." has a bunch of cool performance fixes!
I've upstreamed them into `testing` branch with you as author, so you may need to rebase this to top of testing again.

Regarding the Lua module being slow - did you try something like `cargo-flamegraph` to see if it can point out where the lag occurs?

Edit: I pushed the whole chain to the `testing` branch - did a few cleanups, so individual commits pass the tests and clippy. So you can re-fetch cleanly from `testing` branch head and continue your tuning.

auscyber:
> Your last commit with "close layout gaps..." has a bunch of cool performance fixes! I've upstreamed them into `testing` branch with you as author, so you may need to rebase this to top of testing again.
> 
> Regarding the Lua module being slow - did you try something like `cargo-flamegraph` to see if it can point out where the lag occurs?
> 
> Edit: I pushed the whole chain to the `testing` branch - did a few cleanups, so individual commits pass the tests and clippy. So you can re-fetch cleanly from `testing` branch head and continue your tuning.

i'll ahve a look at cargo flamegraph yes

auscyber:
it ballooned in size but performance was massively taking a hit and i want to go through and make as many of these things use async channels so that they can work properly. also i had to patch sketchybar to allow for dynamic luaPackages to be set so it can use luajit

https://github.com/auscyber/dotfiles/blob/master/packages/sketchybar/sbarlua-external-lua.patch

https://github.com/auscyber/dotfiles/blob/master/packages/sketchybar/_sbarlua.nix
so i would recommend using this to test sketchybar on your end so that you can have it use luajit because the hot paths are quite dense obviously and the performance gains are quite noticeable

auscyber:
im not the biggest fan of having a shared types crate at the moment, but in terms of generating a lua module it seems the only sane way, which we obviously do want a separate lua module for interfacing with sketchybar for example. inherently using a socket like this causes a lot of performance loss so im also gonna look at mach ports as imo they have a lot more performance on macos

auscyber:
unix sockets are a little slow and incoherent so im looking at other ipc methods, especially needed if we this lua api to be fast and able to allow for complex processes (which im sure we do so we can bike shed any feature requests into “write it yourself in lua and if the apis for the lua dont exist add it and thats ur pull request “)
@karinushka do you see it feasible like i do to try and abstract some of the layout management things to be luajit based, so that people can customise it however they want and all that needs to be distributed is lua libraries?

auscyber:
performance seems significantly better with mach ports, especially as the wire protocol was json, and that was really an unpleasant thing to work with, you can still get json from the cli, but you cant just tap into the socket anymore, but i dont think thats fundamentally a nice api anyway, the performance penalty of deserialising and serialising is just unpleasant at best

auscyber:
so i essentially made an async api for mach ports in rust that didnt exist before, i might just publish that as a separate crate and then vendor it in, i implemented it like this, because one didnt exist before, so its worth me publishing it on crates.io

auscyber:
okay this seems to be the best performance i can eek out, have a look, have a test

karinushka:
Update: actually not ☺️
It was me who had a semi-empty `~/.paneru.lua` which naturally did not contain any bindings.

> Something is broken with the first commit: https://github.com/karinushka/paneru/pull/337/changes/d890ff826dd14b6a99e91d04d2ef3abc43abb28c
If I try building Paneru on top of it, it does not respond to keybindings. Previously I thought that it was a bad merge on my side, but now I pulled the clean rebase from your fork and it still does not work?
I assume it works on your side? So something is different somehow? I will try grokking the code and see whether I can spot it.

crates/lua/src/client.rs
```diff
@@ -58,7 +58,10 @@ fn send(argv: &[String]) -> LuaResult<UnixStream> {
 
 /// The primitive the shared API is built on: encode the command and send it to
 /// the daemon (fire-and-forget).
-fn dispatch(_: &Lua, command: &Command) -> LuaResult<bool> {
+// Takes `Command` by value to match the shared `crate::Dispatch` closure type,
+// which every verb closure in `lib.rs` also has to satisfy.
+#[allow(clippy::needless_pass_by_value)]
+fn dispatch(_: &Lua, command: Command) -> LuaResult<bool> {
```
karinushka: Ah, I was reacting to Clippy's suggestion of passing it by reference, since the function was not consuming it. So I changed it to reference everywhere 😺

src/ecs/triggers.rs
```diff
@@ -855,26 +887,34 @@ pub(super) fn window_destroyed_trigger(
...
+        // window back to life, leaving the entity in the strip and a permanent
+        // gap where the window had been.
+        //
+        // `window_ids` rather than `window_list`: the latter reads title, role
+        // and subrole off every window and runs the config rules over them,
+        // which is far too much work for a check that only compares IDs, on a
+        // path that also fires for every window leaving a space.
         if matches!(source, DestroySource::SpaceNotification)
             && window.role().is_ok()
+            && app.window_ids().contains(window_id)
```
karinushka: This breaks windows coming back from native full screen.
It seems there is a short transitional period when a window just exited fullscreen and is converted from a native workspace into an actual window - during that time, it is not reported by the App as its window id's. 🤷

That's why I reshuffled this function and removed the `window_ids()` function for now.

karinushka:
Great stuff! I've cleaned up some commits, added descriptions, etc and pushed to `testing`. Check it out!
The performance fixes are amazing - you went deep diving with the flamegraph 😺

I need to play around with the Lua scripting - but unfortunately I lack imagination on what would be some nifty use-cases. If you have some ideas - capture them in discussion thread or something. We can start a "script tweak" topic where people can share ideas.

auscyber:
> Great stuff! I've cleaned up some commits, added descriptions, etc and pushed to `testing`. Check it out!
> 
> The performance fixes are amazing - you went deep diving with the flamegraph 😺
> 
> 
> 
> I need to play around with the Lua scripting - but unfortunately I lack imagination on what would be some nifty use-cases. If you have some ideas - capture them in discussion thread or something. We can start a "script tweak" topic where people can share ideas.

i straight up no joke went testing a lot produced flamegraphs and then asked claude to restructure based on it. it ended up with some bizarre restructures but they resulted in astronomical reductions in cpu busy time and latency. i moved some onto main thread as condvar was using a lot of cycles. 

in terms of some scripting examples have a look at xmonad contrib

karinushka:
Started a thread: https://github.com/karinushka/paneru/discussions/344

Also I had to rollback the "reap seemingly non-existant windows" - it was erroneously reaping windows on other workspaces during configuration reloads. You probably did not notice it in your testing, because when you change the workspace it auto-detects the windows again. But if you have multiple virtual workspaces, they all disappear.

auscyber:
wait i didnt mean to close this

karinushka:
> wait i didnt mean to close this

No probs - do you want to continue with new commits in this PR or otherwise you can open a new fresh PR with the current `testing` branch as base.

PR #339, opened 2026-08-03
Automatic workspace creation & menu bar/overlay state matching

lun1xr:
# Summary
_These two things were kinda bothering me, so I thought I'd try and add/fix them for QoL. Fair warning I'm not super experienced with Rust, but the changes seem to work well on my local build._

**Virtual Workspaces**
- Altered binding processing for Southward vw switching and Southward window movement to allow for creation of a new vw if one does not exist in that direction
- Gated on the current workspace having windows
- Added automatic workspace creation config option

**Menu Bar Indicator**
- Indicator update pattern now matches how the overlay refreshes, relying on the same dirtying system

## Config
```toml
[options]
create_virtual_workspace_automatically = true
```


CONFIGURATION.md
karinushka, line 1: Your editor seems to have reformatted the whole file and inserted a bunch of spaces. Just add the lines of your documentation 😺

src/ecs/workspace.rs
```diff
@@ -1154,15 +1197,11 @@ fn reap_empty_virtual_workspaces(
...
     for (entity, strip) in rows {
-        if strip.virtual_index > 0 && strip.len() == 0 {
-            if entity == changed_entity {
-                debug!("moving markers from despawned virtual workspace to primary");
-                if let Ok(mut entity_commands) = commands.get_entity(primary_entity) {
-                    entity_commands.try_insert(ActiveWorkspaceMarker);
-                }
-            }
+        if entity != changed_entity && strip.virtual_index > 0 && strip.len() == 0 {
             if let Ok(mut entity_commands) = commands.get_entity(entity) {
```
karinushka, line 1249: These two if statements can be folded into one:
```rust
if entity != changed_entity && strip.virtual_index > 0 && strip.len() == 0
  && let Ok(mut entity_commands) = commands.get_entity(entity) {
```

src/ecs.rs
```diff
@@ -150,11 +150,13 @@ pub fn register_systems(app: &mut bevy::app::App) {
...
                     .after(systems::animate_resize_entities)
                     .run_if(dimming_enabled)
-                    .run_if(overlay_dirty),
+                    .run_if(vw_indicator_dirty),
                 systems::update_flash_messages,
             )
                 .chain(),
-            crate::menubar::update_menu_bar,
+            crate::menubar::update_menu_bar
+                .after(systems::animate_resize_entities)
```
karinushka: Why does it depend on animations? It should be able to run in parallel?

src/ecs/workspace.rs
```diff
@@ -915,13 +940,26 @@ fn move_virtual_workspace_bind(
...
             if active_display.active_strip().len() > 1 =>
         {
-            (current_virtual_index + 1, *move_focus)
+            if current_index + 1 < rows.len() {
+                (rows[current_index + 1].1.virtual_index, *move_focus)
+            } else if config.create_workspace_automatically() {
+                (current_virtual_index + 1, *move_focus)
+            } else {
+                return;
+            }
```
karinushka: This should be:
```suggestion
            if config.create_workspace_automatically() && current_index + 1 < rows.len() {
                (rows[current_index + 1].1.virtual_index, *move_focus)
            } else {
                (current_virtual_index + 1, *move_focus)
            }
```

Otherwise if you don't have the option in your config it prevents normal VW window movements as well.

karinushka (commented):
Can you rebase this branch on top of `testing` instead of `main`? That way we won't inadvertently break the main population ☺️

Also, I tried it out and was wondering about the use-case?
A new VW is already created when you move a window into it (i.e. the `window_virtualmove_south` and `window_virtualsend_south`.
This change allows you to create and switch into an empty workspace. But to move a window into it, you still have to go one step up and use existing bindings. (Also when leaving this empty VW, it may get auto-reaped if setting is enabled).

Is the purpose to provide a visual cue that you reached the end of the VW stack by creating and showing you an empty VW (with floating layer visible).

lun1xr:
Hi, thanks for looking at it. The idea (because this is my preferred workflow), is to be able to scroll to a new workspace and then create windows there rather than operating only on window movement. So for example, I could go to workspace 2 without specifying workspace 2 exactly, and then open whatever I want there.

As for the window movement, maybe it just wasn't working for me or the homebrew version was broken? What I changed in the bind processing which seemed to fix it on my end. If it's working on main, I'll drop it and leave it as a patch just for me.

Will rebase in a minute, I just woke up.

Also sorry about the docs thing, I have no idea why it moved all those other lines. I'll try and edit it again in a regular text editor.

karinushka:
> is to be able to scroll to a new workspace and then create windows there rather than operating only on window movement

Ah! I completely forgot that yeah, you can run new applications on an empty workspace 😺

lun1xr:
Rebased to testing and fixed the config file in nvim, so it shouldn't have any extra spaces. Lmk if I should change anything else (love the app btw, it's singularly put me on to scrolling wms).

src/ecs.rs
lun1xr: Missed this mb; I'm not sure why I did that, probably was thinking of something window related, will fix. 

Looked more closely, I was looking at the mechanism for the overlay, and I copied the line inadvertently.

PR #345, opened 2026-08-09
feat: add installable app launcher

ayamir:
## Motivation

Paneru is currently installed as a CLI/launchd service, so application launchers such as Spotlight, Alfred, and Raycast cannot discover it as an app. Users can wire up custom shell actions, but a small native app bundle provides a simpler and launcher-agnostic entry point.

While testing that flow, I also found that `paneru start` could report success without starting the daemon. The `launchctl` wrapper passed the plist path to `kickstart` and discarded non-zero exit statuses.

## Changes

- add `paneru install-app` and `paneru uninstall-app`
- generate an ad-hoc signed `$HOME/Applications/Paneru.app`
- keep the launcher UI-less with `LSUIElement` and have it invoke the installed Paneru binary with `start`
- stage the bundle next to its destination before replacing it
- refuse to replace or remove an existing app unless its bundle identifier belongs to Paneru
- start bootstrapped launchd services with their service target (`gui/<uid>/<label>`)
- propagate `launchctl` failures instead of logging a false success
- document the app-launcher workflow

## Validation

- `cargo test --locked` — 182 passed
- `cargo clippy --all-targets --locked -- -D warnings`
- `cargo build --release --locked`
- `plutil -lint` on the generated `Info.plist`
- `codesign --verify --deep --strict --verbose=2` on the generated app
- end-to-end macOS test: quit Paneru, open the generated `Paneru.app`, then confirm `paneru query active` and the launchd job return to a running state

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,22 @@ $ paneru install
 $ paneru start
 ```
 
+### Installing an app launcher
+
+To start Paneru from Spotlight, Alfred, Raycast, or another application launcher,
+install the lightweight app wrapper:
+
+```shell
+$ paneru install-app
+```
+
+This creates `$HOME/Applications/Paneru.app`. Opening the app starts the
+installed Paneru launch agent and exits immediately. Remove the wrapper with:
+
+```shell
+$ paneru uninstall-app
+```
+
 ### Running in the foreground
 
 ```shell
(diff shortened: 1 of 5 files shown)


ayamir:
What the maintainer might need to do is create an app icon for paneru.

karinushka:
> What the maintainer might need to do is create an app icon for paneru.

We use this image for the Matrix channel - maybe it can be re-used as an icon?

image

PR #349, opened 2026-08-14
perf: fix bruteforce

auscyber:
make it just a bit faster


auscyber:
the nix doesnt work, you need to make the Cargo.toml not include vendored by default @karinushka

karinushka:
> the nix doesnt work, you need to make the Cargo.toml not include vendored by default @karinushka

I wanted to make it work with normal `cargo build --release`. Can you make Nix run it with custom `--features` flag? E.g. if there is a Lua installed already, Nix can specify it during build?

auscyber:
nix cant actually do that with luajit, its a whole odd thing

karinushka:
Merged it into `testing` - please pull and check that everything works.

I left out the `role` and `subrole` caching - `role` is used to verify that a window is still active during some phantom SLS events: https://github.com/karinushka/paneru/blob/bcf248ef60aa4c0b49002161b97bc6e268b05aa7/src/ecs/triggers.rs#L880

auscyber:
@karinushka i factored out the mach ipc stuff into my own crate, obviously you can audit it. its almost identical.

https://crates.io/crates/async-mach-ports

i just want to use it for other things as well and debloat the repo from this!

also fixed some problems with things like ghostty with native tabs

karinushka:
Yup, pushed your changes to the `testing` branch. Sync up so you use the latest.

karinushka:
I am unsure about the floating window per workspace commit: https://github.com/karinushka/paneru/commit/ccb3c85cd13727dc2cf86e052e2983961e7330cf

Was using it for a while and it changes or breaks a lot of the existing mechanics, the biggest ones are:
- Always on top floating windows from browsers (for example YouTube picture-in-picture).
- Toggle floating layer on/off logic. That one causes the virtual workspace switches to the previous floating layer.

karinushka:
Pushed to `testing`: https://github.com/karinushka/paneru/commit/3aed2f6694c68a347492603bb9bd078920cf2865
Check it out.

PR #350, opened 2026-08-17
fix(input): revive the event tap when macOS invalidates it silently

bensleveritt:
## The problem

macOS can invalidate the event tap's mach port across a sleep cycle without ever delivering `TapDisabledByTimeout` or `TapDisabledByUserInput` to the callback. No further event then reaches the callback, so the re-arm branch there can never fire, and the tap stays dead until the daemon restarts.

Nothing else notices. The IPC channel keeps answering queries and the layout engine keeps tiling on command, so a deaf paneru looks like a healthy one. In practice every keybinding, click and swipe stops working after the machine wakes, with no error anywhere.

## A second, smaller bug

`input_handler` currently reads:

```rust
let Some(events) = &self.events else {
    return false;
};

let result = match event_type {
    CGEventType::TapDisabledByTimeout | CGEventType::TapDisabledByUserInput => { ... }
```

The re-arm sits *below* the early return, so when `events` is `None` the one notice macOS does give is swallowed. This patch moves the re-arm above it.

## The fix

- Split tap setup out of `start()` into `install_tap()` / `uninstall_tap()`, retaining the run loop source so the tap can be torn down and rebuilt in place.
- `ensure_tap_alive()` re-enables a disabled-but-valid port and rebuilds an invalid one, since re-enabling an invalid port is a no-op.
- Drive it from the event loop on `Event::SystemWoke`, which is the usual trigger, plus a 30s sweep to catch the deaths that arrive without one.
- Move the `TapDisabled` re-arm above the early return.

## Notes

Rebased onto `testing` and adapted to the current `pump_events` (the `PUMP_BUDGET` / `try_recv` drain loop) and to `pump_cocoa_event_loop` returning `bool` with the waker re-arm.

`cargo test` on this branch: 271 passed, 1 failed. The failure is `config::test_config_defaults`, which fails identically on clean `testing`.

Happy to adjust the sweep interval or drive it differently if you would rather it hung off something other than `pump_events`.

https://claude.ai/code/session_01RRHjo2QEFc9vorASrZHML7

### DIFF
--- a/src/ecs/systems.rs
+++ b/src/ecs/systems.rs
@@ -37,6 +37,7 @@ use crate::manager::{
     Application, Display, Process, Window, WindowManager, WindowOS, bruteforce_windows,
 };
 use crate::overlay::{FlashMessageManager, OverlayManager};
+use crate::platform::input::TapHealth;
 use crate::platform::{PlatformCallbacks, WinID};
 
 /// Processes and applications still inside their spawn grace period, with the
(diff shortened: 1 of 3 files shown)


GeekDuanLian:
I hope a new version can be released for this, as it's quite affecting usability.

PR #365, opened 2026-08-28
fix: Restore focus when switching virtual workspaces without animations

### LINKED ISSUE #364: Focus not restored when switching virtual workspaces with animations disabled (regression on main)
## Summary

On current `main` (bcf248e), keyboard focus is not restored after switching virtual workspaces when `virtual_workspace_animations = false` (the default). The first directional `focus` command after a workspace switch is consumed by a recovery fallback instead of moving focus, so e.g. `alt+2` followed by `alt+l` (focus east) requires pressing `alt+l` twice. v0.4.4 is not affected.

## Reproduction

Config: `virtual_workspace_animations = false` (default).

1. Row 1: kitty (focused), Zed. Row 2: Dia, WeChat.
2. `paneru send-cmd window virtualnum 2` (or the bound hotkey).
3. `paneru query state` → `focused_*` fields are all `null`.
4. First `paneru send-cmd window focus east` only re-enters the active strip (focus lands on the first window of row 2); only the second `focus east` actually moves east.

With `virtual_workspace_animations = true` the problem does not occur.

## Root cause

Commit `890c85b` ("fix: Raise windows when switching virtual workspaces.") changed the non-animation fallback in `show_active_workspace` (`src/ecs/workspace.rs`, ~1193-1203): v0.4.4's `commands.focus_entity(*entity, false)` (v0.4.4 `workspace.rs:1066-1069`) was replaced by a raise-only `commands.trigger(RaiseWindow { entity: focus, with_strip: true })`. In the non-animation path this no longer sets `FocusedMarker` / keyboard focus, so after switching rows no window is focused.

The animation path still calls `focus_entity`, which is why the bug only shows with animations disabled.

Two fallbacks mask the problem, which makes it look like "one lost keypress":

- `recover_lost_focus` (`src/ecs/focus.rs:356`) doesn't fire because the marker isn't lost — it's still attached to a window on the hidden row.
- The "re-enter active strip" fallback in `src/commands.rs:312-317` consumes the first directional focus command, focusing the first window of the now-active row instead of moving in the requested direction.

## Suggested fix

In the non-animation fallback of `show_active_workspace`, keep the `RaiseWindow` trigger and also restore `commands.focus_entity(focus, false);` — raise first, focus second, since `raise_window_trigger`'s comment requires the target window to be raised last.

A regression test could assert that after a `VirtualNumber` switch with animations disabled, `FocusedMarker` ends up on a window of the target row (the existing cases in `src/tests/interaction.rs` don't cover this path).

## Environment

- paneru: HEAD-bcf248e (built via `brew install --HEAD paneru`)
- macOS: 26.x (Tahoe), Apple Silicon

Hariketsu:
Fixes #364.

## Problem

With `virtual_workspace_animations = false` (the default), switching virtual workspaces does not restore keyboard focus: `FocusedMarker` stays on a window of the hidden workspace, and the first directional `focus` command after the switch is consumed by the re-enter-strip fallback (`src/commands.rs:312`) instead of moving focus. Users observe this as "alt+2 then alt+l needs two presses".

Regression from `890c85b` ("fix: Raise windows when switching virtual workspaces."), which replaced `commands.focus_entity(...)` in the non-animation fallback of `show_active_workspace` with a raise-only `RaiseWindow` trigger. The animation path still calls `focus_entity`, so the bug only shows with animations disabled.

## Fix

In the non-animation fallback of `show_active_workspace`, keep the `RaiseWindow` trigger (it re-orders the strip's windows so the target ends up on top) and additionally restore `commands.focus_entity(focus, false)`. Focus is issued after the raise because `raise_window_trigger` must raise the target window last (raised windows get OS focus events).

## Test

Added `test_virtual_workspace_switch_restores_focus_no_animations` in `src/tests/interaction.rs`: builds a second virtual workspace, visits it, then asserts that switching back and forth with animations disabled immediately restores `FocusedMarker` to each workspace's remembered window. The test fails without the fix and passes with it.

Note: `config::test_config_defaults` fails locally both with and without this change (pre-existing, environment-dependent — it picks up the user config). All other 198 tests pass.

### DIFF
--- a/src/ecs/workspace.rs
+++ b/src/ecs/workspace.rs
@@ -1200,6 +1200,11 @@ pub(crate) fn show_active_workspace(
                 entity: focus,
                 with_strip: true,
             });
+            // RaiseWindow only re-orders the windows; restore keyboard focus
+            // too, otherwise the focus marker stays on a window of the hidden
+            // workspace until the next explicit focus command. Focus after
+            // raising: raise_window_trigger must raise this window last.
+            commands.focus_entity(focus, false);
         }
     }
 }
--- a/src/tests/interaction.rs
+++ b/src/tests/interaction.rs
@@ -13,7 +13,7 @@ use crate::ecs::{
 use crate::ecs::{RepositionMarker, SpawnWindowTrigger};
 use crate::events::Event;
 use crate::manager::{Origin, Size, Window};
-use crate::platform::Modifiers;
+use crate::platform::{Modifiers, WinID};
 use crate::{assert_focused, assert_window_at, assert_window_size};
 
 use super::*;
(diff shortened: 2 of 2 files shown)


Hariketsu:
Closing since `main` already has the fix (713ae49) - thanks for the quick turnaround! 😄

One difference worth noting for the record: this PR calls `focus_entity` **after** the `RaiseWindow` trigger, following the "raise the focused window last" comment in `raise_window_trigger` (raised windows get OS focus events, so focusing after the raise seemed like the safer order). Your commit focuses first, then raises - functionally equivalent as far as I can tell; I've been running a locally patched build with this exact change on macOS 15.7 for two days and the regression is gone.

The regression test here asserts remembered-focus round-trips between two virtual workspaces, but the one you landed covers the same path, so there's nothing left to salvage here. Thanks again!

karinushka:
> One difference worth noting for the record: this PR calls focus_entity after the RaiseWindow trigger

This is a Bevy peculiarity - a trigger is executed immediately after the current function ends, while insertion of a focus marker happens on the next frame (where its triggers will be executed).
So the ordering stays correct.

But now I am wondering why the 'raise window' alone did not change the focus - maybe macOS raise window does not send an application focus event?

PR #367, opened 2026-08-29
fix: retain application observer after window discovery timeout

### LINKED ISSUE #362: paneru refuses to manage Emacs Windows
This is issue #33 which is either back or wasn't fixed completely. If I restart paneru, the window gets managed, otherwise on first or subsequent launch of Emacs it never does.

Emacs Version 31.1 (9.0) on M2 Pro Tahoe 26.6.2

am5083:
Fixes #362.

## Problem

When a newly observed application does not expose an accepted window before the discovery timeout, `timeout_ticker` currently despawns its `Application` entity.

Paneru has already installed the application's accessibility observer at that point. Despawning the entity drops the observer, so windows created later are invisible to Paneru until it is restarted.

## Change

Treat the application timeout as the end of initial window polling:

- remove `FreshMarker` and `Timeout`
- retain the `Application` and its accessibility observer
- preserve the existing despawn behavior for non-application timeouts

This also handles the deferred-command state where `FreshMarker` has already been removed but the expired `Timeout` remains until the next update. The existing five-second discovery deadline remains unchanged.

Polling stops when `FreshMarker` is removed. The retained application and observer are still removed with their parent process when the application terminates.

## Testing

Reproduced on macOS 27 using a regular AppKit application that became observable immediately but created its first standard window after 12 seconds. The unpatched installed Paneru did not discover the late window.

The regression test covers:

- an application timing out with `FreshMarker`
- an application with an expired timeout after marker cleanup
- unchanged despawning of ordinary timed entities
- removal of retained applications when their parent process terminates

Validation on macOS:

- `cargo check`
- `cargo test application_timeout_stops_polling_without_despawning`
- `cargo test -- --skip config::test_config_defaults` — 199 passed
- `cargo clippy --all-targets -- -D warnings`
- `cargo build --release --verbose`

`config::test_config_defaults` is skipped because its macOS-version expectation does not account for macOS 27; this is unrelated to the observer change.

### DIFF
--- a/src/ecs/systems.rs
+++ b/src/ecs/systems.rs
@@ -358,7 +358,7 @@ pub(crate) fn finish_setup(
 /// observes the application for events, and adds its windows to the manager.
 /// This system processes `BProcess` entities marked with `FreshMarker`.
 /// If the process is not yet ready, it continues observing it. If ready, it attempts to create and observe an `Application`.
-/// A `Timeout` is added to the application if it takes too long to become observable.
+/// A `Timeout` stops initial window polling if the application does not expose a window in time.
 ///
 /// # Arguments
 ///
@@ -371,7 +371,7 @@ pub(super) fn add_launched_process(
     config: Res<Config>,
     mut commands: Commands,
 ) {
-    const APP_OBSERVABLE_TIMEOUT_SEC: u64 = 5;
+    const APP_WINDOW_DISCOVERY_TIMEOUT_SEC: u64 = 5;
     let mut already_seen = HashSet::new();
 
     for (entity, mut process, children) in fresh_processes {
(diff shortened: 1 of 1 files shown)


am5083:
The commit is unsigned, not sure if that's a problem.

The bug is fixed (as far as my testing has shown), trying to see if I can get independent verification on another machine/different macOS version, but from what I can tell, this patch should be sufficient.

am5083:
Nevermind, apparently the bug still persists with this patch. I *think* I know the problem, but let me verify and check.

src/ecs/systems.rs
```diff
@@ -484,29 +484,35 @@ pub(super) fn fresh_marker_cleanup(cleanup: TimedOutSpawns, mut commands: Comman
...
 /// # Arguments
 ///
-/// * `timers` - A `Populated` query for `(Entity, &mut Timeout)` components.
+/// * `timers` - Timed entities and a marker identifying applications.
 /// * `clock` - The Bevy `Time` resource for getting the delta time.
-/// * `commands` - Bevy commands to despawn entities.
+/// * `commands` - Bevy commands used to clean up expired entities.
 pub(super) fn timeout_ticker(
-    timers: Populated<(Entity, &mut Timeout)>,
+    timers: Populated<(Entity, &mut Timeout, Has<Application>)>,
```
karinushka, line 498: So this `timeout_ticker` is a generic system which is meant to tick and clean up any component which has a `Timeout` attached.
It's a generalized way to schedule stuff to happen in the future - once the timeout hits, it potentially executes a registered system and then deletes the component.

So adding an application specific logic into it feels wrong.
I am wondering whether the solution is to not attach Timeouts into Applications at all and just keep them spawned.

PR #368, opened 2026-08-29
fix: stop Accessibility stalls from disabling the keyboard event tap

bensleveritt:
## Summary

Hotkeys intermittently stop reaching paneru and land in the focused app instead, and typing feels laggy in between. It reads as a misbehaving keyboard rather than a window manager fault, which is what makes it hard to attribute.

Both symptoms come from the same place. Accessibility getters are synchronous Mach round trips into the target app. paneru makes them on the main run loop, and the `CGEventTap` callback runs on that same run loop. While a call is outstanding the tap cannot fire, so keystrokes are delayed. Hold the run loop long enough and macOS switches the tap off entirely, at which point hotkeys pass through to the focused app until paneru re-arms it.

Three commits, smallest to largest in effect.

## 1. Bound the timeout for window elements too

`ApplicationOS::new` already sets `AXUIElementSetMessagingTimeout` on the application element, with a comment saying that covers every element obtained through it.

`AXUIElement.h` does not document that inheritance. It describes the system-wide element as the object carrying a process-global default, and says an element left at zero "uses the current global timeout value". Window elements are copied out with `AXUIElementCopyAttributeValue` and are never passed to `AXUIElementSetMessagingTimeout`, so on that reading they wait out the six second default rather than the intended 0.25s. Every window attribute read is one of those: role, subrole, title, and the frame reads on the layout path.

This sets the same constant on the system-wide element once at startup. The per-application call is left alone; I did edit the comment beside it, since it now describes only half the picture.

**I have not proven the inheritance question empirically**, only that the documentation does not promise it. If you know the app-element timeout does propagate to copied children, this commit is unnecessary and the other two stand on their own.

## 2. Stop re-reading role and subrole

`is_unknown()` and `is_real()` each fetched both attributes, then the rejection messages fetched them again to build a string the caller only logs at `trace!`. `is_real` now takes the values it already has. A rejected window costs 4 round trips instead of 7.

`window_creation_event` also drained every pending `WindowCreated` in one frame. It now queues them and spends at most 200ms per frame, following the `PUMP_BUDGET` pattern already in that file. Queuing rather than leaving messages unread means nothing is lost if the buffer rotates, and one window always completes per frame so a slow app cannot stall progress.

## 3. Name windows off the run loop

This is the one that mattered most.

`notify_app` called `ax_window_id` to turn a notification's element into a window id. That is `_AXUIElementGetWindow`, a blocking round trip back into the app that just sent the notification, made from the observer callback on the main run loop.

It now retains the element and queues it for a resolver thread, which names the window and forwards the finished event. One thread rather than a task pool, because two focus notifications resolving out of order would leave paneru chasing the wrong window.

## Measurements

Taken on a downstream tree carrying these changes plus local stall instrumentation, on one machine.

Time the main thread spent blocked inside the AX notification callback, from a 60 second `sample`:

| | Blocked per 60s |
|---|---|
| Before | 2.32 s |
| After commit 3 | 0.00 s |

The second figure is structural rather than tuning: the callback no longer appears on the main thread at all.

Stalls over 750ms and tap disables, comparing an active working session before commits 2 and 3 against one after:

| | Before | After |
|---|---|---|
| Window | 33.5 min | 12 h 52 min |
| Stalls | 74 | 29 |
| Tap disables | 14 | 1 |
| Worst stall | 5039 ms | 979 ms |

All 29 remaining stalls fell in the first two hours of active use, and none since, including a morning that produced 18 `kAXErrorCannotComplete` timeouts with no stall and no tap disable. Timeouts without stalls is the intended shape: the cap fails a lookup fast instead of holding the loop.

The instrumentation that produced the stall counts is local and deliberately not part of this PR. It measures the gap between consecutive `pump_events` calls, and it is worth knowing that it cannot see blocking *inside* the pump, which is exactly where commit 3's problem lived.

## Not fixed here

The main thread still parks on Bevy workers doing Accessibility work in `update_passthrough`, `focused_window_id` and window creation. That is what the remaining sub-second stalls are. They sit under the roughly one second the tap tolerates, so they rarely disable it, but the right fix is the same move as commit 3 applied to the layout and focus paths. That is a larger change and belongs on its own.

Upstream `new_with_config` also calls `is_forced_manage` eagerly, which reads the title even for windows it is about to accept. Making that lazy would remove another round trip from the common path, but it reorders the rule lookup so I left it out.

## Test plan

`cargo clippy --all-targets` and `cargo fmt --check` are clean. `cargo test` gives 197 passed and 1 failure, `config::test_config_defaults`, which fails identically on `main` before these commits.

### DIFF
--- a/src/ecs/systems.rs
+++ b/src/ecs/systems.rs
@@ -11,8 +11,9 @@ use bevy::math::IRect;
 use bevy::tasks::AsyncComputeTaskPool;
 use bevy::tasks::futures_lite::future;
 use bevy::time::Time;
+use objc2_core_foundation::CFRetained;
 use objc2_foundation::NSPoint;
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 use std::pin::Pin;
 use std::sync::mpsc::{Receiver, RecvTimeoutError, TryRecvError};
 use std::time::{Duration, Instant};
@@ -38,6 +39,7 @@ use crate::manager::{
 };
 use crate::overlay::{FlashMessageManager, OverlayManager};
 use crate::platform::{PlatformCallbacks, WinID};
+use crate::util::AXUIWrapper;
 
 /// Processes and applications still inside their spawn grace period, with the
 /// `FreshMarker` that says whether the spawn actually completed in time.
@@ -96,6 +98,17 @@ const LOOP_TIMEOUT_STEP: u32 = 1;
 /// leftover events stay in the channel for the next frame to pick up.
 const PUMP_BUDGET: Duration = Duration::from_millis(4);
 const PUMP_MAX_EVENTS: usize = 256;
+
+/// How long [`window_creation_event`] may spend building windows before it has
+/// to hand the frame back.
+///
+/// Building one reads Accessibility attributes, and each read blocks on a round
+/// trip into the owning app. The system runs on a worker the main thread parks
+/// on, and a parked main thread stops servicing the `CFRunLoop`, so a burst
+/// drained in one go starves the `CGEventTap` callback until macOS switches the
+/// tap off. Nothing is dropped when the budget is hit: the remainder waits in a
+/// queue for the next frame.
+const WINDOW_CREATION_BUDGET: Duration = Duration::from_millis(200);
 
 /// Gathers all present displays and spawns them as entities in the Bevy world.
 /// The currently active display (identified by `window_manager.active_display_id()`) is marked with `ActiveDisplayMarker`.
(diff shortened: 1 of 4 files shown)


src/ecs/systems.rs
```diff
@@ -1247,20 +1260,30 @@ pub(crate) fn update_low_power_state(low_power_mode: Option<ResMut<LowPowerMode>
...
+    mut messages: MessageReader<Event>,
+    mut commands: Commands,
+    mut pending: Local<VecDeque<CFRetained<AXUIWrapper>>>,
+) {
+    pending.extend(messages.read().filter_map(|event| match event {
+        Event::WindowCreated { element } => Some(element.clone()),
+        _ => None,
+    }));
+
+    let started = Instant::now();
```
karinushka, line 1273: This optimization is incorrect.

In Bevy, calling a `commands.trigger()` simply queues that closure into a command buffer - which is then executed later (together with other triggers) at a later synchronization point.

So in case of multiple windows spawned - this will queue them all into the command buffer and try to spawn them within the same frame. If this causes stalling of the main thread reading the tap, then we should potentially revert the inlining of the Bevy threads from https://github.com/karinushka/paneru/commit/4696dcddef9a600db0be91fb13b68ca3ee0df4cf. The idea was that running multiple Bevy threads causes some overhead, so running everything in one thread is faster - but maybe this is not the case for other hardware and maybe larger number of windows?

Can you try reverting the https://github.com/karinushka/paneru/commit/4696dcddef9a600db0be91fb13b68ca3ee0df4cf and see whether it fixes your TAP stall issue?

karinushka:
I am wondering whether these TAP stalls are caused by running everything in the same thread (done in https://github.com/karinushka/paneru/commit/4696dcddef9a600db0be91fb13b68ca3ee0df4cf).
Can you try reverting that change in your branch and see whether it fixes all the issues with stalled tap? E.g. then the main thread is kept clean of all the other Bevy executions.

Otherwise the first and the last commit look good!

karinushka:
I pushed the first and last commits to `testing` branch.
I rewrote the window resolving one slightly, to make it cleaner - https://github.com/karinushka/paneru/commit/04f1663b7c08142dadd909fe6451f9888f85a663

Check it out!

PR #373, opened 2026-09-03
fix: keep a manually centered window centered

tiagof:
Hi.

This is a bigger PR. The goal is to fix windows losing "centered" position all the time.
Since several files are modified, please have a thorough review of the changes.

### The problem

Center the focused window (`window center`, e.g. `cmd+alt+g`), then open a new
tab in the browser: the window jumps straight back to where it was. Same loss
when you focus a window on another display and come back to the centered one.

### Reproduce

1. `auto_center = false` (the default) and `[swipe] continuous = false`.
2. Several windows on one workspace, more than a display wide.
3. Focus a browser window, `paneru send-cmd window center`.
4. Open a new tab (`cmd+t`) — the window snaps back to the left edge.
5. Or: focus a window on another display, then focus the centered one again.

### Cause

`window center` centers by repositioning the *strip*, and nothing recorded that
the offset was deliberate. Any later `reshuffle_around` re-derives the offset
from the focused window's frame, and with `auto_center` off and
`continuous_swipe` off the edge invariant then pins the strip to the display's
left edge.

Two paths reach it without any layout having changed:

- A browser opening a tab re-emits an OS focus event for the window that
  already holds focus. `window_focused_trigger`'s `already_focused` branch
  reshuffled on every one of those.
- Returning from another display is a genuine focus change, so
  `autocenter_window_on_focus` runs — and it calls `reshuffle_around`
  unconditionally, even when `auto_center` is off.

### Fix

A repeat focus event now uses `ensure_visible` instead of `reshuffle_around`.
`ensure_visible_in_strip` is already the gentler primitive: a no-op while the
window is fully on screen, otherwise a scroll by exactly the off-screen
shortfall. A duplicate focus event carries no new layout information, so it
should not be able to re-derive the offset — but a window that *has* been
scrolled past an edge is still brought back.

For the paths that still reshuffle, `ManualStripOffset` marks a strip whose
offset the user placed by hand (`center`, `snap`). `reshuffle_layout_strip`
honours the mark for any window that is already fully visible, and drops it when
the window it is asked to expose is off-screen — a placement that hides its own
window is stale.

The mark carries a fingerprint of the strip's shape rather than a change tick:
the animator writes `Position` every frame and `layout_sizes_changed` turns that
into a `LayoutStrip` change, so change detection says "changed" on essentially
every frame. The fingerprint is the ordered column tops with their target
widths, which survives scrolling and tab promotion but not an add, remove,
reorder or resize. Swiping and workspace switches drop the mark too.

Also in here, both found on the way:

- `ensure_visible_in_strip` `return`ed out of its `for` loop in five places
  while consuming each `EnsureVisibleMarker` up front, so the first
  already-visible entity silently swallowed every other request that tick.
  Now `continue`.
- The same function clamped the entity into the viewport on both axes and then
  threw the vertical half away, keeping the strip's own `y`. It now uses the
  clamped `y` as well.

### Tests

New: center survives a repeated focus event; center survives a display
round-trip; center dropped on window add, on workspace switch, on user swipe; a
partly hidden window is still scrolled into view after a center; a unit test for
the fingerprint comparison.

Both halves were verified to carry weight by disabling each in turn — without
`ManualStripOffset` the display round-trip fails, without the `ensure_visible`
swap the repeat-focus case fails.

Full suite: 212 passing, clippy clean.

Generated with Claude Code

### DIFF
--- a/src/ecs/layout.rs
+++ b/src/ecs/layout.rs
@@ -16,7 +16,8 @@ use crate::config::Config;
 use crate::ecs::params::Windows;
 use crate::ecs::{
     ActiveWorkspaceMarker, Bounds, DockPosition, EnsureVisibleMarker, Initializing, LayoutPosition,
-    Position, RepositionMarker, ReshuffleAroundMarker, Scrolling, SpawnCommandsExt,
+    ManualStripOffset, Position, RepositionMarker, ReshuffleAroundMarker, Scrolling,
+    SpawnCommandsExt,
 };
 use crate::errors::{Error, Result};
 use crate::manager::{Display, Origin, Size, Window};
@@ -35,7 +36,8 @@ pub struct LayoutEventsPlugin;
 /// frame has left the display entirely to whichever display it landed on.
 pub(crate) const PARKED_STRIP_SLIVER: i32 = 10;
 
-/// A strip, its entity, origin, display, and whether it's the active one.
+/// A strip, its entity, origin, display, whether it's the active one, and the
+/// offset it is animating toward if a move is already in flight.
 /// Shared by [`reshuffle_layout_strip`] and [`ensure_visible_in_strip`].
 type StripPlacements<'w, 's> = Query<
     'w,
@@ -46,6 +48,7 @@ type StripPlacements<'w, 's> = Query<
         &'static Position,
         &'static ChildOf,
         Option<Ref<'static, ActiveWorkspaceMarker>>,
+        Option<&'static RepositionMarker>,
     ),
 >;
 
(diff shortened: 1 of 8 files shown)


src/ecs/layout.rs
```diff
@@ -1019,10 +1055,31 @@ fn layout_strip_changed(
     }
 }
 
+/// Drops a [`ManualStripOffset`] once the strip it was taken on no longer has
+/// the same shape — a window added, removed, reordered or resized. The offset
+/// only ever meant "the user placed *this* layout here".
+#[instrument(level = Level::DEBUG, skip_all)]
+fn invalidate_manual_offsets(
+    strips: Query<(Entity, &LayoutStrip, &ManualStripOffset)>,
```
karinushka: Use `Populated` instead of `Query`. This will not schedule the system if the pattern doesn't match. A small optimization which is missed out by LLMs for some reason.

karinushka:
Yup, this looks good!
A small nit, just like in the other PR - get the LLM to write the unit tests in the "standard" format instead of manually operating on the `World` entity. E.g.:
```rust
    let harness = TestHarness::new().with_config(config).with_windows(5);
    let commands = vec![
        Event::MenuOpened { window_id: 0 },
        // ...
     ];
    harness
        .on_iteration(1, |world, _state| {
              // do checks
        })
        .run(commands);
```

Edit: also, since this is a larger change - rebase it on top of `testing`. That way we can bake it for a bit before unleashing onto the general `main` population.

tiagof:
Thanks @karinushka for taking the time to go through my multiple PRs. As I mentioned in previous PRs, i'm not familiar at all with rust, so these are AI generated.
Still, i love using paneru and not knowing a specific language is, these days, irrelevant.

Will work on this and the other PRs as requested!

Cheers

tiagof:
Rebased on `testing` and rewrote the tests in the harness `run(commands)` + `on_iteration` pattern — including the display round-trip one in `tests/display.rs`.

One thing worth flagging from the rebase. The first version invalidated a stale `ManualStripOffset` from a small system of its own in the layout chain. On `testing` that broke ~24 unrelated tests, and it turned out not to be my system's logic: adding *any* system to `Update` does it. A no-op `fn noop_probe_system() {}` registered on top of a clean `testing` fails 23-24 tests on its own — window/strip positions come out different, e.g. `test_window_shuffle` and `test_scrolling`. It looks like the schedule has systems whose relative order is unconstrained, and the executor's order for them shifts when the graph changes, so those tests are effectively pinned to today's ordering. Worth a look independently of this PR — a new system anywhere is going to hit it.

The fix here no longer adds a system: the signature is checked where the mark is read, in `reshuffle_layout_strip`, which is the only place that consults it. Full suite is green on `testing` (293 passed).

Unrelated: `src/ecs/layout_ops.rs` on `testing` is not `cargo fmt` clean. Left it alone to keep this diff to the point.

Written with Claude Code.

src/ecs/layout.rs
tiagof: Good rule — noted. This particular system is gone in the rebased version, though: `invalidate_manual_offsets` was what broke ~24 unrelated tests on `testing`, and the staleness check now happens where the mark is read, inside `reshuffle_layout_strip`.

The one query the PR still adds is `manual_offsets: Query<&ManualStripOffset>` in that system, and there `Populated` would be wrong: it is a lookup by entity (`manual_offsets.get(strip_entity)`), not the pattern the system iterates. Gating on it would stop `reshuffle_layout_strip` running whenever no strip carries a manual offset, which is nearly always.

So: `Populated` when the query is the system's work, `Query` when it is an auxiliary lookup.

karinushka:
Ooh, very interesting about the tests being flaky due to random order of the systems - I'll look into it!

karinushka:
The flaky tests should be fixed in https://github.com/karinushka/paneru/commit/5cb8a9fb0282a922f1a30dab66d6e712ee07f25c
It was about system ordering.

PR #356, opened 2026-08-22
feat: make menubar configurable

lun1xr:
# Notes
Added a bunch of config options for the menubar. See updates to CONFIGURATION.md as it's a lot to write again. Authored mostly by me with some Opus 5 revisions. I've reviewed everything and done a fair amount of testing.
## Example config
```toml
[decorations.menu]
orientation = "default"
colors = ["#FF0000","#00FF00"]
angle = 90

[decorations.menu.descriptor]
style = "symbol"
symbol = "fish"
text = "VW"

[decorations.menu.indicator]
style = "multi"
format = "unicode"
font_size = 13
active_character = "X"
inactive_character = "O"
```
## Demos
Example config:
image

Both/Default:
image

Hidden/Roman:
image

My config (that I think is cute because it looks like a fish and bubbles):
image

PR #374, opened 2026-09-04
feat: add paged and marked menubar workspace indicators

tiagof:
Two additions to `[decorations.menu.indicator]`, on top of #356.

`style = 'paged'` — the active workspace over the total, `2 / 4`. It stays one glanceable width however many workspaces exist, which `multi` does not. The total is formatted from the last workspace's index, so it goes through the same formatter as every other position and reads `IV` under `format = 'roman'`.

`format = 'marked'` — `unicode` with the active workspace kept as its number: `· 2 · ·`. The marks say how many workspaces there are and where you sit among them; the number says which one that is, which a fixed active character cannot.

Both only mean anything beside the workspaces they are not, so the single-item styles (`mono`, `paged`) fall back to numbers, as they already did for `unicode`.

Written with Claude Code.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -122,12 +122,12 @@ Settings for the descriptor component of the menubar display. The descriptor com
 ### `[decorations.menu.indicator]`
 Settings for the indicator component of the menubar display. The indicator component shows your current (and optionally all active) virtual workspace(s).
 
-**Important Note: The `'unicode'` format can only differentiate between active and inactive workspaces, so it is incompatible with the `'mono'` style and will fall back to `'default'`.**
+**Important Note: The `'unicode'` and `'marked'` formats only mean anything next to the workspaces they are not, so they are incompatible with the single-item `'mono'` and `'paged'` styles and will fall back to `'default'`.**
 
 | Option | Type | Default | Description |
 | :--- | :--- | :--- | :--- |
-| `style` | String | `'mono'` | Whether the indicator displays all virtual workspaces `'multi'`, or only your current workspace `'mono'`. |
-| `format` | String | `'default'` | The format of the character(s) which represent your virtual workspace(s).<br>Options are regular numbers `'default'`, Roman numerals `'roman'`, or any pair of Unicode characters `'unicode'`. |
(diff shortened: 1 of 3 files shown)


karinushka:
Looks good!
Can you rebase this on top of `testing` branch, because there is already a PR which does some menubar adjustments (https://github.com/karinushka/paneru/pull/356) so it would be good to rebase this on top of it.

tiagof:
Rebased on `testing`, and #356 turned out to cover nearly all of what this PR was doing, so I dropped my version rather than land a second config surface for the same thing:

| this PR, before | #356 |
| :--- | :--- |
| `workspace_menu_style = 'single'` | `indicator.style = 'mono'` |
| `'dotted'` | `indicator.style = 'multi'` + `format = 'unicode'` |
| `'full'` | `indicator.style = 'multi'` + `format = 'default'` |
| `workspace_menu_icon = false` | `descriptor.style = 'hidden'` |
| `workspace_menu_font_size` | `indicator.font_size` |

What was left is `'paged'` (`2 / 4`), so that is all this PR is now: one `IndicatorStyle` variant, the render arm, docs, and two unit tests. The label formatting moved into a free `indicator_label` function on the way, which is what made it testable without a menu bar.

Two differences I did not carry over, in case you want them later: my `dotted` kept the active workspace's *number* between the dots rather than swapping it for a character, and `full` drew the active number knocked out of a filled rounded rect (a template image, so the system tints it in either menu bar). Both are cosmetic variants of `multi` and would fit as extra formats if they seem worth it.

Suite is green on `testing` (288 passed).

Written with Claude Code.

tiagof:
Added `format = 'marked'` alongside `'paged'`: the inactive character for the others, the active workspace as its number — `· 2 · ·`. That was the one visual from the original version of this PR that #356's config could not express (`unicode` swaps the active workspace for a fixed character, `default` numbers all of them), and it is the layout I use daily.

It is a match arm in `indicator_label` plus the same single-item fallback `unicode` already has. The active number renders bold, following what `default` and `roman` do for the active position.

289 passed on `testing`.

Written with Claude Code.

PR #375, opened 2026-09-04
feat: copy a window rule from the menu bar

tiagof:
Writing a `[windows]` rule means knowing the app's bundle id and the window's
exact title. Neither is visible anywhere in Paneru's UI, and the only way to
read them today is `paneru send-cmd printstate` and squinting at the log.

Adds a **Copy Window Rule** entry to the menu bar. It puts a rule for the
currently focused window on the clipboard:

```toml
[windows.ghostty]
bundle_id = "com.mitchellh.ghostty"
title = "^paneru — zsh$"
# title = ".*"   # all windows of this app
# app: Ghostty  role: AXWindow  subrole: AXStandardWindow
# floating = true
# manage = true
# width = 0.5
```

Only `title` and `bundle_id` are matchers in `find_window_properties`, so those
are the two live lines; the rest is commented. `title` is `regex::escape`d and
anchored, since `is_match` is unanchored and an unescaped title with a `.` or
`(` in it would either fail to compile or match the wrong windows — the
commented `.*` covers the "every window of this app" case instead. The `app` /
`role` / `subrole` comment is only there to tell one pasted rule from the next;
none of it is matchable.

When an `init.lua` has taken the TOML file out of play (`CONFIGURATION_FILE`
is `None`), the snippet is emitted as a Lua table instead.

### Notes

- `Operation::CopyRule` is issued only by the menu bar, following the
  `Command::Lua` / `Command::Layout` precedent. `Operation::to_argv` now
  returns `Option<Vec<String>>` so it can say "no argv encoding" the same way
  `Command::to_argv` already does.
- `NSPasteboard` use is isolated in `src/pasteboard.rs`, per AGENTS.md §2
  (no `objc2` in ECS systems). No `Cargo.toml` change: `objc2-app-kit` is
  already pulled in with default features, which cover `NSPasteboard`.
- Snippet generation is a pure function in `src/config/snippet.rs`. The escaping
  order is the part that can silently go wrong — `regex::escape` adds
  backslashes that then need escaping again for the config string — so one test
  parses the generated TOML back through `InnerConfig` and asserts the rule
  matches the very title it was generated from.
- The new system is registered separately because the existing `PreUpdate` tuple
  was already at Bevy's 20-system limit.

### Testing

`cargo fmt`, `cargo clippy --all-targets` clean; `cargo test` (212) and
`cargo test --features lua,vendored,luajit` (285) pass. Verified by hand on a
release build: focused a window, clicked the entry, pasted the rule back into
`paneru.toml` and confirmed it takes effect.

🤖 Generated with Claude Code

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -297,6 +297,30 @@ horizontal_padding = 5
 bindings_passthrough = ["ctrl-h", "ctrl-l"]
 ```
 
+### Copying a rule from the menu bar
+
+Neither the bundle ID nor the exact window title is visible anywhere in the UI,
+so the Paneru menu bar has a **Copy Window Rule** entry. It puts a rule for the
+currently focused window on the clipboard, ready to paste:
+
+```toml
+[windows.ghostty]
+bundle_id = "com.mitchellh.ghostty"
+title = "^paneru — zsh$"
+# title = ".*"   # all windows of this app
+# app: Ghostty  role: AXWindow  subrole: AXStandardWindow
+# floating = true
+# manage = true
+# width = 0.5
+```
+
+The two matchers are live; everything else is a comment for you to uncomment.
+`title` is anchored and regex-escaped so it matches only this window — swap in
+the commented `.*` to cover every window of the app instead. The `app`, `role`
+and `subrole` line is there to tell one pasted rule from the next; none of those
+are matchable. When an `init.lua` is in charge, the snippet is written as a Lua
+table instead of TOML.
+
 ### Forcing management of LSUIElement or non-standard windows
 
 Some applications (e.g., BetterTouchTool, ProtonVPN) are flagged as background apps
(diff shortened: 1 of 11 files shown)


karinushka:
There is also a `query` command which prints the window state as JSON ☺️
But yeah, this is a very cool feature! Much easier to just click on the menu than manually copying from a terminal!

Just like with the other PRs, can you rebase it on top of `testing` to give some baking time before hitting the `main` branch.

tiagof:
Rebased on `testing`; it applied cleanly and the suite is green there (293 passed).

Good point about `query` — it is what I used to check the output while building this, and it stays the tool when you want the whole state. The menu entry is aimed at the other half of the job: it hands you a `[windows]` block you can paste, with the title already regex-escaped and anchored (a title with `(`, `+` or `.` in it is easy to get wrong by hand), the non-matcher fields as comments so the rule stays valid, and a Lua table instead of TOML when an `init.lua` has taken the TOML file out of play.

Written with Claude Code.

karinushka:
I did a small tweak in a follow up change: https://github.com/karinushka/paneru/commit/98f057b82a5ad3ee271d0ef9a046b1cc290dee60

Lua does not like fields starting with digit, so I changed "1password" to "onepassword" automatically. Also the option of copying a window rule to clipboard is quite nifty - so having it as a keybinding as well would be nice. Now you can both pick it in a menu and as a keyboard shortcut 😺

PR #377, opened 2026-09-05
fix: focus and expose the window a Cmd-Tab switched to

tiagof:
Cmd-Tab to an app whose window sits on a hidden virtual workspace switches to that workspace, but the window you picked does not end up focused and visible.

**Focus.** The activation is caused by the focus event itself, so the window already holds `FocusedMarker` by the time `show_active_workspace` runs. Its animated branch then restored the strip's remembered focus unconditionally — the non-animated branch was guarded, the animated one was not — so focus jumped to whatever was focused on that workspace last. The guard now covers both.

**Visibility.** The strip was restored to the offset it had when it was parked, which says nothing about the window that just took focus: it could land almost entirely off-screen. The later reshuffle derives the offset from the window's own frame, which mid-restore is still parked, so the window often settled only partly visible. The restore now aims at an offset that shows the whole window, and marks it for a re-check on the next frame in case the layout is still settling. `ensure_visible_in_strip` projects against the strip's target offset rather than its current position, so that re-check is not read off a strip still in motion.

Only a focus that arrives with the activation moves the strip — a window that held focus all along (switching back from an empty workspace, say) keeps the offset it was left at.

Tests: two harness tests covering both halves, plus a unit test for the new `origin_exposing` helper.

Written with Claude Code.

### DIFF
--- a/src/ecs/layout.rs
+++ b/src/ecs/layout.rs
@@ -138,6 +138,19 @@ fn signature_matches(current: &[(Entity, i32)], stored: &[(Entity, i32)]) -> boo
         })
 }
 
+/// The strip offset that shows all of a window sitting at `layout` within the
+/// strip, starting from `origin` and moving no further than the window's
+/// shortfall past a viewport edge. Returns `origin` unchanged when the window
+/// already fits.
+pub(crate) fn origin_exposing(
+    layout: Origin,
+    size: Size,
+    origin: Origin,
+    viewport: IRect,
+) -> Origin {
+    clamp_origin_to_viewport(layout + origin, size, viewport) - layout
+}
+
 impl Plugin for LayoutEventsPlugin {
     fn build(&self, app: &mut App) {
         app.add_systems(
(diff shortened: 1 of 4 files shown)


karinushka:
Thanks for catching these regressions - I was not using workspace switch animations for some time, so I did not notice the focus breakage.

Nit: ask the LLM to re-write the unit test using existing Harness pattern, where it sets up the events and then does checks on each iteration:
```rust
    let harness = TestHarness::new().with_config(config).with_windows(5);
    let commands = vec![
        Event::MenuOpened { window_id: 0 },
        // ...
     ];
    harness
        .on_iteration(1, |world, _state| {
              // do checks
        })
        .run(commands);
```

Most of the unit tests use this pattern and it makes it easier to understand what is going on. A few non-conforming tests managed to get in and now they continuously keep confusing the LLMs 😺

Edit: also, since this is a larger change - rebase it on top of `testing`. That way we can bake it for a bit before unleashing onto the general `main` population.

tiagof:
Rebased on `testing` (so it sits on top of #373) and rewrote both tests in the harness `run(commands)` + `on_iteration` pattern.

One mock change came out of that. `focus_without_raise` was a no-op, so after Paneru moved focus internally the app kept reporting its previous window — and `window_focused_trigger` discards an event whose window is not what the app says is focused. The stale focus event for the Cmd-Tab target then arrived *after* the wrong restore, put focus back, and healed the bug a frame later, in the test only. The mock now moves the app's focused window there too, without emitting the notification a real focus change would produce; the whole suite stays green with it (296 passed).

The exposure test leans on `window_hidden_ratio = 0.5`: a reshuffle is allowed to leave a window that far over an edge, which is what makes the restore's own aim the thing under test rather than the correction that follows it.

Written with Claude Code.

PR #379, opened 2026-09-09
feat: focus stack neighbor before switching virtual workspace (window_virtualfocus_north/south)

soraxas:
## Summary

This is similar to `focus-window-or-workspace-up` and `focus-window-or-workspace-down` in niri.

Adds a new command, `window_virtualfocus_north`/`window_virtualfocus_south` (`Operation::FocusOrVirtual`): if the focused window is in a stack and has a neighbor above/below, it focuses that neighbor — the same within-column traversal `window_focus_north`/`_south` already uses. Only when there's nothing left to focus that way does it switch the virtual workspace, exactly like `window_virtual_north`/`_south`.

This is deliberately a **separate command**, not a change to `Operation::Virtual`'s existing behavior. An earlier version of this change made `window_virtual_north`/`_south` themselves do the focus-first check, which silently changes what an already-established command means for anyone using it purely as a workspace switch. `window_virtual_north`/`_south` keep their original, unconditional "switch workspace" meaning; `FocusOrVirtual` composes with it internally (it reduces to a plain `Operation::Virtual(direction)` once the focus check doesn't apply), so the substantial index-cycling logic in `switch_virtual_workspace_bind` is shared, not duplicated.

No `_east`/`_west`/`_first`/`_last` form is added for the new command — there's no "focus" reading of those directions to pair with a workspace switch.

## Test plan

- [x] `cargo test` — full suite green
- [x] New test: focused stack member cycles to its sibling via `FocusOrVirtual(South)` without switching workspace; a second press (nothing left to focus) falls through to switching
- [x] New test: `FocusOrVirtual(North)` with no stack switches workspace immediately, same as plain `Virtual`
- [x] Regression test: plain `Operation::Virtual(South)` still switches workspace unconditionally even when the focused window has a stack sibling available — pins the separate-command decision so it can't quietly erode back into the coupled behavior
- [x] `cargo clippy` clean, `cargo fmt` applied

## Example config

```toml
[bindings]
window_virtualfocus_north = "alt - e"
window_virtualfocus_south = "alt - n"
```


karinushka:
Can you resync this and other PR's to head of `testing` again. Publishing `v0.5.0` caused some branch changes.

karinushka:
I pushed this to the current `testing`: https://github.com/karinushka/paneru/commit/334dd5bcec71b56172d50708ebc9210f8d70f27d
Resync and check it out.

soraxas:
Yep it works well!

PR #380, opened 2026-09-09
fix: stop stack members from sliding on a virtual workspace restore

soraxas:
## Summary

`virtual_workspace_animations = false` only guaranteed the strip's own aggregate offset restored instantly. Two separate gaps let individual member windows visibly animate into place anyway, one tick later — most noticeably a window in a stacked column appearing to "reveal from its corner" on switch.

**Bug 1** — `show_active_workspace` defers "expose the arriving focus window" to `ensure_visible_in_strip`, because that system's own `is_added(ActiveWorkspaceMarker)` guard skips it on the activation tick itself. By the time it actually runs (one tick later), `is_added` is no longer true, so it had no way to distinguish "standing in for the restore" from an ordinary reshuffle correction, and always animated regardless of the config flag.

**Fix**: `EnsureVisibleMarker` now carries a `snap: bool`. `show_active_workspace` sets it to `!virtual_workspace_animations()`; every other caller (ordinary, non-restore corrections) keeps `false`.

**Bug 2** — `position_layout_windows` infers "was this move restore-driven?" purely from distance: an offscreen-magnitude heuristic (80% of viewport height), or being at the parked corner. A lower stack member parked while its strip was hidden can land at a position just under *both* thresholds — close enough that the heuristic reads the correction as an ordinary small layout change and animates it, even though it's really the tail end of the restore.

**Fix**: a new `SnapStripMarker` guard, spawned by `show_active_workspace` at the exact point it writes the restored strip's `Position` directly. While a live guard names a strip, `position_layout_windows` snaps every member of that strip unconditionally, instead of guessing from move magnitude.

Root-caused from a live `RUST_LOG=paneru=trace` capture, not guesswork: the exact numbers (a stack member's actual Y sitting under both the offscreen-distance and parked-corner thresholds, by a comfortable margin) matched the observed `animate_entities` slide precisely.

## Test plan

- [x] `cargo test` — full suite green
- [x] New regression test: an `EnsureVisibleMarker { snap: true }` correction never animates with `virtual_workspace_animations = false`; a companion test confirms the ordinary (non-restore) path still animates as before
- [x] New regression test: a strip named by a live `SnapStripMarker` snaps an under-threshold-perturbed member directly instead of animating; a companion test confirms the same perturbation *does* animate without the guard, proving the scenario is a real repro and the guard isn't a no-op
- [x] Confirmed both fixes are load-bearing by disabling each in turn and watching its regression test fail, then restoring it
- [x] `cargo clippy` clean, `cargo fmt` applied


karinushka:
Pushed this to the `testing` branch: https://github.com/karinushka/paneru/commit/a4d87f61e307e5415496e0db51b5f9ac1110a036
Please sync and check it out.

soraxas:
Yep it works well! thanks!

PR #381, opened 2026-09-09
feat: niri-style tabbed display for stacked columns

### LINKED ISSUE #302: Tabs support
https://niri-wm.github.io/niri/Tabs.html would be a really neat feature to emulate!

soraxas:
## Summary

https://github.com/user-attachments/assets/d549e22c-6bf1-40cb-84ae-0fe9fe903f73

closes #302 

Adds niri's "tabbed column" (https://niri-wm.github.io/niri/Tabs.html): a `Column::Stack` can toggle to a tabbed display — one window visible at a time, sharing the full column size — cycled with the existing `window_focus_north`/`_south` keys. New command `window_tabbeddisplay` toggles it on/off, with a flash-message confirmation ("Tabs on"/"Tabs off").

Deliberately **does not touch** `Column::Tabs`/`StackItem::Tabs`. That variant is strictly an inference layer over macOS's own native "Merge All Windows" tabbing (geometry + focus-event detection), also exposed to Lua as `ColumnKind::Tabs` — paneru never creates/hides/resizes those members itself, the OS does. Conflating the two would break real invariants other code depends on (`apply_window_positions` skips resizing tabbed entities assuming the OS owns their frame; `give_away_focus` assumes the OS auto-focuses the remaining native tab). This is a genuinely separate, paneru-driven concept:

- **Data model**: a `tabbed_stacks: EntityHashSet` membership set on `LayoutStrip` — a Stack is "tabbed" iff its members are in the set (kept all-or-none by `toggle_tabbed_display` and cleaned up on collapse/removal/merge). No `Column` enum shape change, so no ripple through the ~15 other exhaustive matches on `Column`.
- **Visibility**: pure z-order overlap — every tabbed member gets the identical full-column frame in `relative_positions`, and whichever one is raised (by the normal focus-raise path) is the only one visible. No new hide/show choreography, no new rendering.
- **Cycling**: reuses `Operation::Focus` North/South unchanged — `get_window_in_direction`'s stack traversal already had no idea whether a column was tabbed, so it just worked.
- **Real bug fix along the way**: `give_away_focus` needed a genuine fix — closing the *active* tab used to fall through to a nearest-column search that can't see other tabs in the same column (its own entry was the one just closed), incorrectly jumping focus to an unrelated column. Now checks for a same-column sibling first.

**Explicitly out of scope for v1** (each independently addable later, no rework needed): the clickable tab-indicator strip, a config surface, Lua exposure, session-restore persistence, and fullscreen-with-multiple-tabs. 

## Example config

```toml
[bindings]
window_tabbeddisplay = "alt - w"
```


karinushka:
Pushed this to the `testing` branch: https://github.com/karinushka/paneru/commit/9897d03d4dd91dea7fe34efc253160a2b0780960
Check it out whether it works as expected.

One thing I noticed is that it doesn't remember the leader change - e.g. if window A and B are stacked, changing to B, going to another column and returning back would focus A again.

PR #386, opened 2026-09-10
fix: keep virtual workspace row 0 instead of renumbering a space from "2"

tiagof:
## The bug

A native space can end up holding virtual rows 1 and 2 with no row 0. Its
menu bar indicator then reads `2 3`, and nothing can get back to workspace
"1" — `virtualnum 1`, `virtualmovenum 1` and `virtualsendnum 1` are all
no-ops, and northward navigation walks the rows that exist, so it stops at
row 1.

Live example from a v0.5.0 daemon, on the space whose indicator read `2 3`:

```
$ paneru query state --json
vw {'native_workspace_id': 1, 'number': 1} nwin 0   <- padding, no strip behind it
vw {'native_workspace_id': 1, 'number': 2} nwin 3
vw {'native_workspace_id': 1, 'number': 3} nwin 1
```

## Where it came from

Two paths despawn an empty row 0:

- `find_orphaned_workspaces` despawns any empty orphaned strip when a display
  goes away — `09f5394` ("fix: Reparent workspaces both on add and move of a
  display", v0.3.7). `789e158` later moved the code into `ecs/workspace.rs`.
- `restore_window_state` despawns an existing strip whose windows were
  consumed onto the restored rows, which is exactly what happens to row 0 when
  the saved state keeps its windows on other rows — #223 ("feat: Restore window
  sessions on startup", v0.4.3).

Nothing renumbers the remainder: `renumber_virtual_indexes` only resolves
*duplicate* indexes among newly added strips, and
`reap_empty_virtual_workspaces` already refuses to reap row 0. So the hole is
permanent once it appears.

What makes it unrecoverable instead of self-healing came in #204 ("feat(api):
add JSON state and absolute workspaces", v0.4.2): the numbered switch and move
binds both special-case index 0 and `return` instead of spawning the missing
strip, making it the one index that does not auto-create. The same PR added the
`1..=max` padding in the query document, which hides the symptom — a bar built
on `paneru query` renders a slot for workspace 1 that no command can select,
while the menu bar lists only the rows that exist.

## The fix

Treat row 0 as the invariant it already is (`workspace_created_handler` creates
every space with it):

- `find_orphaned_workspaces`: re-parent an empty row 0 like a populated strip
  rather than despawning it.
- `restore_window_state`: keep and reuse row 0 when restore empties it.
- `switch_virtual_workspace_bind` / `move_virtual_workspace_bind`: drop the
  index-0 bail so a numbered target spawns row 0 like any other index.
  `Direction::First` already targeted index 0 unconditionally and let
  `handle_virtual_window_moves` spawn the destination, so no new machinery was
  needed.

Deliberately unchanged: the query document still pads missing numbers, since
middle-row gaps remain legitimate (reaping an empty row 1 of 0/1/2 leaves one)
and integrations rely on stable numbered slots.

## Tests

Four regression tests, one per path (312 pass, `cargo clippy --all-targets`
clean):

- `test_empty_baseline_row_survives_display_removal` — empty row 0 is orphaned
  and re-parented across a display removal/return instead of despawned.
- `test_startup_restore_keeps_emptied_baseline_row` — restore leaves row 0 in
  place next to the restored row 1.
- `test_virtual_number_recreates_missing_baseline_row` — `virtualnum 0`
  recreates row 0 and makes it active.
- `test_virtual_move_number_recreates_missing_baseline_row` — a numbered move
  lands the followed window on a recreated row 0.

`test_multi_workspace_orphaning` also becomes meaningful: its assertions used
to iterate an empty set, because both empty strips had been despawned.

Assisted by Claude Code.

### DIFF
--- a/src/ecs/restore.rs
+++ b/src/ecs/restore.rs
@@ -465,7 +465,11 @@ pub(super) fn restore_window_state(
             strip.remove(*entity);
         }
 
-        if had_consumed_window && strip.all_windows().is_empty() {
+        // Row 0 is the row every space is created with. Emptying it is normal
+        // (its windows move onto the restored rows), but despawning it leaves
+        // the space numbered from "2" with no row that any later switch or
+        // reap path recreates, so it is kept and reused instead.
+        if had_consumed_window && strip.all_windows().is_empty() && strip.virtual_index > 0 {
             emptied_existing_strips.insert(entity);
         }
     }
--- a/src/ecs/workspace.rs
+++ b/src/ecs/workspace.rs
@@ -484,7 +484,10 @@ fn find_orphaned_workspaces(
     let present = window_manager.present_displays();
 
     for (orphan, orphan_entity, timeout, child) in orphans {
-        if orphan.len() == 0 {
+        // Row 0 is the row every space is created with, so it is re-parented
+        // like a populated strip: despawning it leaves the space numbered from
+        // "2" with no row that any switch or reap path recreates.
+        if orphan.len() == 0 && orphan.virtual_index > 0 {
             if let Ok(mut cmd) = commands.get_entity(orphan_entity) {
                 cmd.try_despawn();
             }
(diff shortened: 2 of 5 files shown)


tiagof:
@karinushka tagging you to bring your attention to what seems a relatively important fix for a bug in v0.5.

PR #389, opened 2026-09-10
fix: size the menubar status item to include its content padding

### LINKED ISSUE #387: Menubar item off-center when descriptor is hidden: content inset MENU_BAR_SPACING but setLength(width) ignores it
## Summary

With the menubar descriptor hidden (`decorations.menu.descriptor.style = "hidden"`), the workspace number is **not centered inside the status item's hover region** — it is shifted right and slightly clipped at the right edge. The layout appears to be hardcoded rather than configurable.

## Repro

```toml
[decorations.menu.descriptor]
style = 'hidden'
```

(any indicator style; most visible with `mono` + a single digit)

**Expected:** the number sits centered in the hover/clickable area of the menubar item.

**Actual:** the number is inset 5pt from the left of the hover region and overflows its right edge by the same amount.

## Root cause (from `src/menubar.rs` on current `main`)

`install()` frames the content with a constant leading inset, but sizes the status item to the content width only:

```rust
const MENU_BAR_SPACING: CGFloat = 5.0;
...
view.setFrame(CGRect::new(
    CGPoint::new(MENU_BAR_SPACING, origin_y),   // content starts 5pt in from the left
    CGSize::new(width, height),
));
button.addSubview(view);
self.status_item.setLength(width);              // hover region = content width only
```

So the content occupies `[5, 5 + width]` inside a hover region of `[0, width]` — off-center to the right, with the right 5pt spilling past the item boundary.

With the default fish descriptor visible this is barely noticeable (wide item, two elements). Hiding the descriptor leaves a single digit in a tight item, where a constant 5pt offset is a large fraction of the width — that's when it becomes obvious.

## Suggested fix

Account for the inset on both sides:

```rust
self.status_item.setLength(width + 2.0 * MENU_BAR_SPACING);
```

(or center the frame: `x = (item_length - width) / 2.0` with `item_length = width + 2 * MENU_BAR_SPACING`).

## Environment

- paneru 0.5.0 (brew), macOS 15
- Lua config, descriptor `hidden`, indicator `mono`/`default`

tiagof:
Fixes #387.

## The bug

`install` frames the content at `MENU_BAR_SPACING` but sizes the status item to
the bare content width:

```rust
view.setFrame(CGRect::new(
    CGPoint::new(MENU_BAR_SPACING, origin_y),
    CGSize::new(width, height),
));
self.status_item.setLength(width);
```

So the content occupies `[5, 5 + width]` inside an item of `[0, width]` — one
padding to the right of centre, with the trailing 5pt clipped past the item
boundary. As the issue notes, the default fish descriptor masks it (wide item,
small relative offset); `descriptor.style = 'hidden'` with a `mono` indicator
leaves a single digit in a tight item, where it is obvious.

## The fix

Reserve the padding on both sides, which is the issue's own suggestion. The
content origin and the item length are computed together in
`status_item_metrics`, so the centring invariant is stated once rather than
split across two call sites that have to agree:

```rust
fn status_item_metrics(content_width: CGFloat) -> (CGFloat, CGFloat) {
    let length = content_width + 2.0 * MENU_BAR_SPACING;
    ((length - content_width) / 2.0, length)
}
```

The content's origin works out to the same 5pt it already used, so the only
behavioural change is the item being 10pt wider — the trailing padding it was
already drawing into now belongs to it.

Untouched: the `width <= 0.0` early return still sets a zero length (no padding
around nothing), and the gradient path is unaffected since the content frame's
*size* does not change, only the item around it.

## Test

`status_item_reserves_padding_on_both_sides_of_the_content` in the existing
`menubar::tests` module asserts the content is centred in the item and keeps
its leading padding, across several content widths including the
single-digit case. 309 pass, `cargo clippy --all-targets` clean.

Not visually verified against a `descriptor = 'hidden'` setup — the geometry is
the arithmetic above, but a screenshot from the reporter would be worth having
before merge.

Assisted by Claude Code.

### DIFF
--- a/src/menubar.rs
+++ b/src/menubar.rs
@@ -432,15 +432,16 @@ impl MenuBarManager {
             return true;
         }
 
+        let (origin_x, length) = status_item_metrics(width);
         let origin_y = ((self.status_bar.thickness() - height) / 2.0).max(0.0);
         view.setTranslatesAutoresizingMaskIntoConstraints(true);
         view.setFrame(CGRect::new(
-            CGPoint::new(MENU_BAR_SPACING, origin_y),
+            CGPoint::new(origin_x, origin_y),
             CGSize::new(width, height),
         ));
         button.addSubview(view);
         button.setToolTip(Some(&NSString::from_str("Paneru window manager")));
-        self.status_item.setLength(width);
+        self.status_item.setLength(length);
         true
     }
 
@@ -625,6 +626,16 @@ pub fn update_menu_bar(
     );
 }
 
+/// Where the content sits inside the status item, and how wide that item has
+/// to be: [`MENU_BAR_SPACING`] of padding on both sides of the content. Sizing
+/// the item to the bare content width instead leaves the content pushed right
+/// by one padding and clipped by the same amount on the other side - obvious
+/// once the descriptor is hidden and the item is only a digit wide.
+fn status_item_metrics(content_width: CGFloat) -> (CGFloat, CGFloat) {
+    let length = content_width + 2.0 * MENU_BAR_SPACING;
+    ((length - content_width) / 2.0, length)
+}
+
 pub(crate) fn virtual_workspace_label(virtual_index: u32) -> String {
     (virtual_index + 1).to_string()
 }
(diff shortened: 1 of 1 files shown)


tiagof:
Superseded by #390, which fixes #387 from the other end.

That PR stops handing the status item button a live view and gives it a rendered image instead, because the live view and AppKit's menu bar replicant snapshots keep re-dirtying each other — a loop that costs about three quarters of a core and starves the run loop the event tap is serviced on, so keybindings die every 30 to 60 seconds. Once the button holds an image, the item can go back to `NSVariableStatusItemLength` and AppKit measures it and applies the item's own content insets: 30pt of content in a 46pt item, 8pt a side, centred. That is the padding this PR computes by hand, so `status_item_metrics` and its test go away there.

The two are alternatives, not a stack. Leaving this open rather than closing it — if #390 is not the direction you want, or needs work, this remains the smaller fix against the current implementation. Happy to close it as soon as #390 lands.

Assisted by Claude Code.

PR #390, opened 2026-09-10
fix: stop the menu bar item from starving the event tap

tiagof:
## Problem

Paneru sits at roughly three quarters of a core while idle, and after a while
every keybinding stops working — the keystroke goes to the focused app instead,
with this in the log every 30 to 60 seconds:

```
WARN paneru::platform::input: event tap was disabled, re-enabled it
WARN paneru::ecs::systems: input tap not healthy (woke=false), recovery: Reenabled
WARN paneru::platform::input: event tap was unusable, rebuilt it
WARN paneru::ecs::systems: input tap not healthy (woke=false), recovery: Rebuilt
```

The two are the same bug. `sample` on the daemon, main thread, ~1000 samples:

```
412  __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__
     __NSFireDelayedPerform
     -[NSStatusItem _updateReplicantsUnlessMenuIsTracking]
     -[NSStatusItem _redrawReplicantSnapshot:sourceView:]
     -[NSView(CacheDisplay) _cacheDisplayInRect:toBitmapImageRep:]
     -[CALayer renderInContext:]                       ...

443  __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__
     CA::Transaction::flush_as_runloop_observer(bool)
     -[NSStatusBarButtonCell drawWithFrame:inView:]
     -[NSSceneStatusItem _setSelectedContentFrame:options:]
     _CASCreateFencePort -> mach_msg   (blocking on the WindowServer)
```

`install` gives the status item button a live view as a subview — an
`NSStackView` of labels, or an `NSView` hosting a `CAGradientLayer` masked by a
bitmap of them. AppKit snapshots the button into its menu bar replicants, and
snapshotting that view re-dirties it: the replicant update schedules a redraw,
the redraw sets the item's content frame, and setting the frame schedules
another replicant update. Nothing converges.

The event tap's callback is serviced on the same run loop, so it stops being
serviced in time. macOS switches the tap off, `ensure_tap_alive` finds a tap
that will not re-enable and rebuilds it, and the fresh tap dies to the same
starvation on the next sweep.

## Fix

Draw an image instead of hosting a view. The content is rendered once, off
screen, into the bitmap the button takes as its image, so no live view is left
in the menu bar to invalidate itself and the snapshot has nothing to feed.

Three things follow from it:

- The image is a template, so AppKit tints it for the current menu bar
  appearance instead of leaving the labels in whatever colour they were
  rendered in — the descriptor symbol is no longer white on a light menu bar.
- Gradients are composited into the bitmap through `NSGraphicsContext` and
  punched out with `DestinationIn`, which is what the mask layer was doing.
  `NSGradient` takes the angle directly, so `unit_point_for_angle` goes with
  the layer that needed it.
- Sizing returns to `NSVariableStatusItemLength`. AppKit measures the image and
  applies the item's own content insets, which is the padding `install` was
  reserving by hand — the same off-centre item reported in #387, fixed by not
  computing it.

This supersedes #389, which fixed that padding arithmetically on top of the
subview. Close it in favour of this if you agree with the approach.

## Test

304 pass, `cargo clippy --all-targets` clean, `cargo fmt --check` clean. Three
tests went with the code they covered: the two `unit_point_for_angle` groups and
`status_item_reserves_padding_on_both_sides_of_the_content`.

Verified on the daemon, my own config (`descriptor.style = 'hidden'`, `multi`
indicator, `marked` format), macOS 26.0:

- idle CPU ~75% before, ~0.6% after
- no tap warnings since, where the previous build logged a death every 30 to 60
  seconds
- the item draws the same as before

The render itself was checked separately against a windowless view, a flat
stack and the nested stack `show_workspaces` builds — 185 painted pixels in each
case, so `cacheDisplayInRect:` off screen loses nothing.

Assisted by Claude Code.

### DIFF
--- a/src/menubar.rs
+++ b/src/menubar.rs
@@ -1,19 +1,18 @@
 use bevy::ecs::query::{Has, With};
 use bevy::ecs::system::{NonSendMut, Query, Res};
 use objc2::rc::Retained;
-use objc2::runtime::AnyObject;
 use objc2::{
     AnyThread, DefinedClass, MainThreadMarker, MainThreadOnly, define_class, msg_send, sel,
 };
 use objc2_app_kit::{
-    NSBitmapImageRep, NSColor, NSControlStateValueOff, NSControlStateValueOn,
-    NSDeviceRGBColorSpace, NSFont, NSImage, NSImageScaling, NSImageView, NSLayoutAttribute, NSMenu,
-    NSMenuItem, NSScreen, NSStackView, NSStatusBar, NSStatusItem, NSTextField,
+    NSBezierPath, NSBitmapImageRep, NSCellImagePosition, NSColor, NSCompositingOperation,
+    NSControlStateValueOff, NSControlStateValueOn, NSDeviceRGBColorSpace, NSFont, NSGradient,
+    NSGraphicsContext, NSImage, NSImageScaling, NSImageView, NSLayoutAttribute, NSMenu, NSMenuItem,
+    NSScreen, NSStackView, NSStatusBar, NSStatusItem, NSTextField,
     NSUserInterfaceLayoutOrientation, NSVariableStatusItemLength, NSView,
 };
 use objc2_core_foundation::{CGFloat, CGPoint, CGRect, CGSize};
 use objc2_foundation::{NSArray, NSInteger, NSObject, NSString};
-use objc2_quartz_core::{CAGradientLayer, CALayer};
 use tracing::warn;
 
 use crate::accessibility_prompt::{AccessibilitySetupAction, show_accessibility_setup};
@@ -307,7 +306,7 @@ impl MenuBarManager {
         field.sizeToFit();
         let size = field.frame().size;
 
-        if self.install(&field, size.width, size.height) {
+        if self.install(self.content_image(&field, size).as_deref()) {
             self.current_content = Some(content);
         }
     }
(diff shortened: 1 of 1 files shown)


tiagof:
One behavioural difference worth calling out, found while running this on a two-display setup.

Before this change, each display's menu bar appeared to show *its own* workspace number. That was never implemented — `update_menu_bar` builds the content from `active_display.active_strip()`, a single value, and there is one `NSStatusItem`. What produced it was the bug: macOS mirrors a status item onto every display's menu bar as replicants, which are snapshots of the button. With a live subview the real view can only be in one menu bar window at a time, so the other display's replicant was a stale capture from when that display was last active — which happened to be that display's last workspace. The redraw loop kept re-capturing, always a step behind.

With a flat image, AppKit propagates the same current image to every replicant, so all displays now show the active display's workspace. That is what the code has always said it does.

No public `NSStatusItem` API gives a status item different content per display, so restoring the per-display reading would mean drawing our own borderless windows in each screen's menu bar strip — a separate feature, and one that has to guess at placement around the notch and Control Center. Left alone here; the workspace popup already announces switches per display.

Assisted by Claude Code.

PR #395, opened 2026-09-12
Fix window resize lag after toggling the Dock or menu bar auto-hide

lmarkmann:
These changes introduce a requery of the usable screen area upon toggling, instead of waiting for a delayed window-size refresh. Additionally, I added a few short follow-up checks while the frame resizes and settles within the new frame size. These changes keep full-width sizing and floating window placement intact as the accessible space for the window adjusts. They additionally correct the menu bar inset, which was computed from the wrong bounds.

Validation: Clippy and release build pass. Dock toggling confirmed locally.


karinushka:
Can you rebase this to `testing` branch. This way we can bake it a bit more before unleashing onto the `main` population.

Additionally, there is a commit in `testing` (https://github.com/karinushka/paneru/commit/38f641e958a306bc288f508cc3f5fb42c2d72dfc) which simply verifies a window size on focus - this works great during display changes and removes the need for all that complex window refresh size logic. So this should reduce the size of this PR.

lmarkmann:
> Can you rebase this to `testing` branch. This way we can bake it a bit more before unleashing onto the `main` population.
> 
> 
> 
> Additionally, there is a commit in `testing` (https://github.com/karinushka/paneru/commit/38f641e958a306bc288f508cc3f5fb42c2d72dfc) which simply verifies a window size on focus - this works great during display changes and removes the need for all that complex window refresh size logic. So this should reduce the size of this PR.

Will do!

lmarkmann:
Rebased onto `testing`.

Dropping the refresh logic worked out well. The PR is down to 2 commits, +240/-12 from +550/-54.

PR #396, opened 2026-09-12
feat: Issue a stable signing certificate so the accessibility grant survives updates

lmarkmann:
Rebuilding / updating paneru leaves the system settings > accessibility showing paneru as enabled, while it actually doesn't work yet. This is due to the ad-hoc signing process and each of the program's updates resulting in a new non-linking cdhash H"...". It results in the dance of removing paneru, then restarting it and enabling it again within the settings.

On the other hand, signing against a stable certificate puts the requirement on the certificate instead, which doesn't change when the binary does, and therefore doesn't reprompt on updates. Therefore, I have done two commits aiming to fix this:
- The first moves the signing into one place, pinning the signed identifier and lets PANERU_CODESIGN_IDENTITY pick an identity. This would only help people that already know how to create a certificate; and deliberatly issue one for paneru.
- The second one therefore issues one on first installation, which is then reused by paneru from then on out.

The certificate is thereby issued only once and never has to rotate.

Verification: locally tested on macOS 26.6.2 with two launches which have distinct cdhashes and the same certificate. The second installation resulted into managing the window layout automatically without having to regrant accessibility settings and the same dance described above. This commit doesn't add any new dependencies.


karinushka:
How come you closed this? Does it not work or?

PR #399, opened 2026-09-16
fix: handle device-independent scroll modifiers

shrirambalaji:
Some scroll events contain a macOS modifier flag without a left/right key flag. Paneru missed those modifiers, which broke Alt + scroll with Deskflow.

Read the general flag when the side is absent. Keep explicit left/right bindings strict.

The modifier tests pass. Full checks still hit the existing macOS 27 border-radius test failure and three Clippy duration warnings.


karinushka:
Pushed this to `testing` branch: https://github.com/karinushka/paneru/commit/65832378b0ab1ca6861b66cab09e9b4044d0dd35
Please check it out.

PR #400, opened 2026-09-16
feat: add optional scroll-wheel window steps

shrirambalaji:
Add an option to select one adjacent column with modifier + mouse wheel:

```toml
[swipe.scroll]
window_step = true
```

Use the existing focus commands, limit repeats to 180 ms, and ignore momentum. The option is off by default. The existing scroll path stays in place.

Tests cover direction, defaults, repeat timing, momentum, and strip edges. Full checks still hit the existing macOS 27 border-radius test failure and three Clippy duration warnings.


karinushka:
Thanks a lot! Actually quite a nifty feature - going to use it myself 😺

Pushed it to `testing` branch with a few cleanups: https://github.com/karinushka/paneru/commit/f79b66ca34c217af1dbf779c9e2fd7a475277c99
Please check it out.

